PR #72, opened 2014-07-15
Input parsing formats

SylvainPlessis:
Here is the chemkin parser in the new parsing design.
There are two/three levels in this design:
  - the read_reaction_data, templated around a parser
  - the parser itself,
  - if needed, low-level methods for the parsers (tiny_xml for the xml parser, none needed for chemkin)

The global high-level method is somewhat clumsy, a little bit too much adapted to the xml parser, nothing really serious, but I keep an option to make it more elegant and standardize a bit more the parsers.
A noteworthy thing is that though it's not paranoid (yet), it already won't let a parser get away with a bad unit or bad representation, this is a good point for any parser wanted in the future, there is no need to know in details what's happening to have something that works cleanly.

The chemkin parser is not exhaustive, but it's fully operable with the file I have from combustion (added in the inputs for the test) and others I've seen  Basically you'll have ``standard'' kinetics with it.  I've finally choose the option to write it almost from scratch, please have a look at the *ascii_getline* method (src/utilities/include/antioch/string_utils.h).  It's basically a *getline* method that deals with the different end-of-line a file can contain, and clearly the chemkin file I had started its life on Windows.  This is something to be expected I guess.  Anyway, I want at the very least an approval of this method before anything happens.

Also it lacks a little bit of documentation, stating for instance that no photochemistry is supported by the chemkin parser, that the weird SRI falloff is not supported either, stuffs like that we might want to deal with in the futur.


SylvainPlessis:
Specifics, this class stores the specificities of the ChemKin format.  What name could be nice and descriptive?  I'm not super good for that...

pbauman:
> What name could be nice and descriptive? I'm not super good for that...

Me neither. `ChemKinParsingHelper`?

SylvainPlessis:
It's not so much a helper than the definitions specific to ChemKin, which just gave me the idea of ```ChemKinDefinitions```, what do you think?

pbauman:
:+1:

SylvainPlessis:
There you go!

pbauman:
OK, so GCC is happy.

I'm good to merge this, but what are the restrictions on the ChemKin parsing?

SylvainPlessis:
Restrictions about what?  By definition it's a hard-coded no suppleness 80 characters by line format, only a Kooij model is supported by this format, and the bastard three-body falloff chemical process is supported in the corresponding far away PR (#98).

pbauman:
Sorry. What I meant was are there aspects of the ChemKin format that we aren't supporting yet? Or can we feed any old ChemKin file and Antioch will work?

SylvainPlessis:
There is the custom intervals and temperatures that is not (yet) supported, where in the input file you define those intervals and the transitions temperatures.

There is the aberration of the Three Body Falloff model (support in PR #98), a rare falloff model, and any custom rule used locally on a chemkin-derived file.

Then there may be others that eluded me, but once the three-body mating inconsiderately with the falloff will be merged, only weirdos won't run (meaning a great majority of the chemkin input file should be ok).

pbauman:
Cool, so the biodiesel thing Jed posted above will work with this?

SylvainPlessis:
> Cool, so the biodiesel thing Jed posted above will work with this?

It should, I haven't tested it, but I used this file for the weird ChemKin case that we should support (and cried at every step of it...).  But I cannot guarantee that within those forty-twelve billions lines no nasty surprise is lurking, waiting for a nice thermo-kinetics library to temper with.

SylvainPlessis:
And note also that you need the falloff three body bastard (PR #98) to make it run as expected.  At first I though it was a mistake or a convenience to switch from falloff to three-body (it actually returns an error), but it turned out to be a blaspheme we should support.

Issue #58, opened 2014-03-12
Rethink ParticleFlux Handling?


pbauman:
I admit ignorance in what the ParticleFlux stuff is doing, but it is potentially taking away thread safety.

So *Mixture classes are intended to be "program constant" classes while *Evaluators are (intended to be) lightweight objects that are cheap to construct (from the *Mixture classes) so that we construct *Mixtures once and *Evaluators many times (e.g. once a thread is forked). However, at least in the case of Planet-INC/Planet_INC, the particle flux pointer is getting reset many times during the program lifetime and, in particular, will be reset by multiple threads at once leading to thread safety problems.

Is there anyway we can move the ParticleFlux business up to the KineticsEvaluator level? It looks like the ParticleFlux is being used to augment the reaction rate - can't we just do that in the KineticsEvaluator when the rate is actually used?

SylvainPlessis:
I guess now this one is obsolete too, isn't it?

SylvainPlessis:
Obsolete, closing it

PR #74, opened 2014-08-21
Conditions reac

SylvainPlessis:
This is the branch that proposes a new API to deal with differences in chemical conditions and solves the thread safety problem (issue #58), right now only temperature and particle flux(es) are managed, generalization to multi-temperature, catalysis, any other stuff is easy by design.

It's quite big but there's (not so?) much to it.  

The global view is that instead of passing down a _const StateType & T_ from _KineticsEvaluator_ to the kinetics models (Arrhenius, Kooij and consorts), it passes down a _KineticsConditions_ object, which provides to the kinetics rate the right conditions (temperature or particle flux), the fork being made at the _KineticsType_ level, the base object to the rate constant calculators.
This _KineticsConditions_ object stores pointers to the conditions, either provided by the user himself
```
ChosenScalarType T = chosen_value_of_temperature;
ParticleFlux<ChosenVectorType> pf(wavelengths,irradiance);

... stuff to know what is the index of the photo reaction ...

KineticsConditions<ChosenScalarType,ChosenVectorType> kin_cond(T);
kin_cond.add_particle_flux(pf,index_of_concerned_photo_reaction);
```
or automatically generated by the _KineticsEvaluator_ to ensure backward compatibility.  I've found a not too expensive way: it either copies a reference or initializes the _KineticsConditions_ object.  Thus the changes in the _kinetics_evaluator.h_ file.
the meta funny structure _constructor_or_reference_ does very exactly what its name implies: if the two templates arguments are the same, it sends back a reference to the type, if not it sends back the type:
```
template<typename T1, typename T2>
struct constructor_or_reference
{
   typedef T1 type;
};

template <typename T>
struct constructor_or_reference<T,T>
 {
   typedef T & type;
 };
```

The biggest part of the changes are the photochemistry handling, in particular the rebinning phase.  Things to be aware are the new utilities functions _conjunction_,  _disjunction_ and _eval_index_, I expect them to be slow for anything other than Eigen, I haven't found a proper method for VexCl and metaphysicL (maybe I haven't been looking hard enough, I'm not sure).

This rebinning part is fully vectorized, sadly slow and I still consider as an almost bug that it recomputes the rate constant whatever happens, even if it does not change.  Good news is I have a few ideas playing with the _KineticsConditions_, though it needs better testing.

As photochemistry is not really essential right now, I think it's ok to proceed.

### DIFF
--- a/src/core/include/antioch/kinetics_conditions.h
+++ b/src/core/include/antioch/kinetics_conditions.h
@@ -1,0 +1,131 @@
+//-----------------------------------------------------------------------bl-
+//--------------------------------------------------------------------------
+//-----------------------------------------------------------------------el-
+
+#ifndef ANTIOCH_KINETICS_CONDITIONS_H
+#define ANTIOCH_KINETICS_CONDITIONS_H
+
+// Antioch
+#include "antioch/antioch_asserts.h"
+#include "antioch/particle_flux.h"
+
+// C++
+#include <vector>
+#include <map>
+
+namespace Antioch{
+
+  /*! This class contains the conditions of the chemistry
+
+      Conditions are:
+        - Temperature(s)
+        - particle flux
+        - any other not yet implemented conditions (catalysis for instance)
+
+      Pressure is given by the molecular composition of the mixture,
+      we are in a gaz phase, so the state equation will give pressure (typically
+      ideal gas).
+
+      The idea is to avoid any copy of anything as much as we can, 
+      we store pointers but deal only with references. Nothing
+      belongs to this object, if there is any cleaning to do,
+      it must be done elsewhere.
+
+      Might be interesting to see about the "double" case
+      for temperature: faster to copy instead of copying
+      the reference, and it's expected to be the classic
+      case.
+   */
+  template <typename StateType, 
+            typename VectorStateType = std::vector<StateType> >
+  class KineticsConditions
+  {
+        public:
+
+          KineticsConditions();
+          KineticsConditions(const StateType & temperature);
+          ~KineticsConditions();
+
+          void add_particle_flux(const ParticleFlux<VectorStateType> & pf, unsigned int nr);
+
+          void change_temperature(const StateType & temperature);
+
+          void change_particle_flux(const ParticleFlux<VectorStateType> & pf, unsigned int nr);
+
+          const StateType & T() const;
+
+          const ParticleFlux<VectorStateType> & particle_flux(int nr) const;
+
+        private:
+        // pointer's not const, temperature is
+          StateType const * _temperature; 
+        // pointer's not const, particle flux is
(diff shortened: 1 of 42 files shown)


src/core/include/antioch/kinetics_conditions.h
```diff
@@ -0,0 +1,175 @@
...
+      the reference, and it's expected to be the classic
+      case.
+   */
+  template <typename StateType, 
+            typename VectorStateType = std::vector<StateType> >
+  class KineticsConditions
+  {
+        public:
+
+          KineticsConditions();
```
pbauman: Do we ever want to do this? I'd make this private and force the user to provide a temperature to construct it.

src/core/include/antioch/kinetics_conditions.h
```diff
@@ -0,0 +1,175 @@
...
+          KineticsConditions(const KineticsConditions<StateType,VectorStateType>& rhs);
```
pbauman: Why's the default copy constructor not enough?

src/core/include/antioch/kinetics_conditions.h
```diff
@@ -0,0 +1,175 @@
...
+
+          const StateType & T() const;
+
+          const ParticleFlux<VectorStateType> & particle_flux(int nr) const;
+
+          const std::map<unsigned int, ParticleFlux<VectorStateType> const *> & map_pf() const;
+
+        private:
+        // pointer's not const, temperature is
+          StateType const * _temperature;
```
pbauman: I'd store a `const StateType&` if we can require the user to supply the temp.

SylvainPlessis:
The copy constructor is a legacy of my ignorance of defaults for those.  I need to dig up a little bit to remember why the default constructor is not private (though the primal reason may have disappeared), and this is a pointer and not a reference to enable temperature change without rebuilding the whole thing.

roystgnr:
Internal compiler error from gcc 4.6?  I'll see if I can reproduce it locally.

roystgnr:
Sorry, that was supposed to be a comment on the rebased PR, #102

pbauman:
Micro version might be important. IIRC, GCC 4.6.2+ is needed with VexCL?

roystgnr:
We're using 4.6.3 on Travis.

ddemidov:
Weren't vexcl tests disabled on Travis-CI due to memory limitations?

SylvainPlessis:
Not on this branch.  We're working on cleaning this PR madness...

roystgnr:
Yeah, this looks like memory limitations with gcc 4.6.  On my local 4.6.3 and 4.6.4 gcc versions, the tests do appear to all pass eventually, but the compiler sucks up 4GB in the process.

roystgnr:
Rebased versions merged via #102

PR #88, opened 2014-12-05
Species parsing

SylvainPlessis:
graph
### Common part
```
xml_parser <-- input_parsing_format <-- ascii_parser <-- species_parsing * <-- \
                                                                                \
NASA polynomials <-- UnifiedThermo <---------------------------------------------- thermo_parsing <--
transport_new <------------------------  transport_thermo <--------------------------------------------- transport_thermo_parser
```
```almost_Exact``` takes in the ```kin_reset```.

### This PR

The species list is parseable by everyone.  The main changes is certainly the names of the files, there is no more reason now to keep the _ascii_ in the name, this is the biggest change.


SylvainPlessis:
Alright updated to change templates to base class, it should be ok.  PR #111 should be no more needed.

roystgnr:
You're missing parser_base.h in pkginclude_HEADERS in src/Makefile.am

roystgnr:
Functions in headers need to either be declared inline or inside class definitions (thus implicitly inline).  Otherwise it's easy to end up with lots of duplicate symbols at link time, as I'm seeing now.  I'll prepare a PR to pull into your branch.

roystgnr:
Gah.  I was all in favor of making antioch_version() inline to make us header-only, but I didn't notice that we'd also made our TinyXML2 import into a header.  That's going too far IMHO.  I won't futz with it now though.

roystgnr:
This passes all my tests now.  Ready to merge if @pbauman is happy.

pbauman:
Looks good!

PR #89, opened 2014-12-05
Unified thermo

SylvainPlessis:
graph
### Common part
```
xml_parser <-- input_parsing_format <-- ascii_parser   <-- species_parsing <-- \
                                                                                \
NASA polynomials <-- UnifiedThermo * <-------------------------------------------- thermo_parsing <--
transport_new <------------------------  transport_thermo <--------------------------------------------- transport_thermo_parser
```
```almost_Exact``` takes in the ```kin_reset```.

### This PR

This is the (not definite) answer to the thermodynamics lack of elegance.  This is the "look mom I did my bed" answer: it looks clean from the outside, don't you look under the blanket.
There is with this PR two new objects: 
 - ```ThermoEvaluator<CoeffType, ExternalThermo,InternalThermo>```
 - ```IdealGasInternalThermo<ExternalThermo,CoeffType>```

The denomination _External_ and _Internal_ are for the macro and micro respectively description of the thermodynamics, thus Cp, h, s, g for the "external" and the vibrational, rotational, translational for the "internal".
The evaluator object is but a reference holder for those two descriptions, and can performs the most common operations, thus no need to go down to the required thermo object.
The ideal gas thermo is the defaulted statistical thermodynamics objects.  It is basically a simplified ```StatMechThermo```.  It requires a macro thermo calculator to obtain Cv_vib with the formulae
```Cv_vib = Cp - R - Cv_rot_trans```, thus a Cp calculation is required.

The structure is always the same Evaluator <-- Mixture <-- stuffs, with "<--" meaning "contains a reference to".  There is no ```ThermoMixture``` as the internal and external thermo are supposed to be in the scheme Evaluator, Mixture, stuff if necessary.  It is not true for the internal part as we have only one option, but is certainly true for the external part.
So it holds a ```NASAEvaluator<CoeffType,NASAFit>``` and either a ```IdealGasInternalThermo<NASAEvaluator<CoeffType,NASAFit>,CoeffType >``` or a ```StatMechThermodynamics<CoeffType>```.

### DIFF
--- a/src/thermo/include/antioch/ideal_gas_internal_thermo.h
+++ b/src/thermo/include/antioch/ideal_gas_internal_thermo.h
@@ -1,5 +1,26 @@
 //-----------------------------------------------------------------------bl-
 //--------------------------------------------------------------------------
+//
+// Antioch - A Gas Dynamics Thermochemistry Library
+//
+// Copyright (C) 2014 Paul T. Bauman, Benjamin S. Kirk, Sylvain Plessis,
+//                    Roy H. Stonger
+// Copyright (C) 2013 The PECOS Development Team
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the Version 2.1 GNU Lesser General
+// Public License as published by the Free Software Foundation.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc. 51 Franklin Street, Fifth Floor,
+// Boston, MA  02110-1301  USA
+//
 //-----------------------------------------------------------------------el-
 
 #ifndef ANTIOCH_IDEAL_GAS_INTERNAL_THERMO_H
--- a/src/thermo/include/antioch/ideal_gas_internal_thermo.h
+++ b/src/thermo/include/antioch/ideal_gas_internal_thermo.h
@@ -94,21 +94,21 @@ namespace Antioch
 
   template <typename ExternalThermo, typename CoeffType>
   inline
-  const CoeffType IdealGasInternalThermo<ExternalThermo, CoeffType>::cv_tr_over_R(unsigned int s)
+  const CoeffType IdealGasInternalThermo<ExternalThermo, CoeffType>::cv_tr_over_R(unsigned int s) const
   {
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by SylvainPlessis** [comment]
rebased

**Comment by pbauman** [comment]
Why introduce `ThemoEvaluator`? Is this going to be needed in subsequent PRs?

**Comment by pbauman** [comment]
Also, I think I might prefer `MacroThermo` and `MircoThermo` as opposed to `ExternalThermo` and `InternalThermo`, respectively, particularly since you make this mapping in the PR comment?

File: src/thermo/include/antioch/ideal_gas_internal_thermo.h [hunk 1]
```diff
@@ -0,0 +1,147 @@
...
+{
+
+  template <typename CoeffType>
+  class ChemicalMixture;
+
+  template <typename CoeffType>
+  class TempCache;
+
+
+  template <typename ExternalThermo, typename CoeffType = double>
```
pbauman: Why does `InternalThermo` care about `ExternalThermo`?

SylvainPlessis:
The Macro/Micro terminology is better, I agree.  I don't remember the history of the External/Internal terminology.

It's needed after, the transport part needs access to both thermo, it simplifies greatly the access and use of it to have only one instance to cope with both micro and macro thermo quantities.

pbauman:
> The Macro/Micro terminology is better, I agree. I don't remember the history of the External/Internal terminology.

OK, feel free to update the PR with this unless @roystgnr objects.

> It's needed after, the transport part needs access to both thermo, it simplifies greatly the access and use of it to have only one instance to cope with both micro and macro thermo quantities.

OK, that's fair. But could we rename the object? This doesn't really fit our `Mixture/Evaluator` paradigm and confused me originally. Maybe `ThermoHandler`?

src/thermo/include/antioch/ideal_gas_internal_thermo.h
SylvainPlessis: To obtain Cv_vib, when you don't have access to the vibrational levels, you use the relationship between Cv and Cp, you get Cv, then you substract Cv_tr, and there you have Cv_vib.

MacroThermo provides Cp.

SylvainPlessis:
I though it was close enough for the ```Evaluator``` name.  ```Handler``` is nice.

src/thermo/include/antioch/ideal_gas_internal_thermo.h
pbauman: Ah, OK, sorry I missed that. Damn. Yeah, I don't see how to separate that here.

roystgnr:
Looks okay to me.  VexCL+MetaPhysicL+GRINS+FIN-S are happy too.  Ready to merge if you both agree.

pbauman:
:+1:

PR #92, opened 2014-12-05
Transport thermo parser

SylvainPlessis:
graph
### Common part
```
xml_parser <-- input_parsing_format <-- ascii_parser   <-- species_parsing <-- \
                                                                                \
NASA polynomials <-- UnifiedThermo <---------------------------------------------- thermo_parsing <--
transport_new <------------------------  transport_thermo <--------------------------------------------- transport_thermo_parser*
```
```almost_Exact``` takes in the ```kin_reset```.

### This PR

Well the title says it all, everything is now happy together.

I'm not sure if we need to indeed merge all the pull requests, I don't fully remember how much changes it required, only that it's less and less.   I'm still leaving the PR though, if it becomes unnecessary because nothing more than compatibility was encoded, deleting it should be easy.


pbauman:
#117 was merged making this obsolete.

PR #95, opened 2014-12-09
Computations for thermodynamics for kin do

SylvainPlessis:
graph
not necessitate to loop over all the species,
only reactants and products are concerned.

This is the commit message, nothing more needs really to be said.

It is fully compatible with the other PR, as it changes a file that is not touched by the others.


roystgnr:
Superceded by #106

PR #98, opened 2014-12-12
Falloff three body

SylvainPlessis:
This one is branched off the **kinetics parsing** PR, it is not mergeable with the **Conditions reac** branch, and any daughter branch (thus **opt transport** and **opt kin on transport**).

Because it starts to be difficult to understand where what goes in what order, here is a more beautiful version (and full) of all the PRs.
graph

Thus to be merged last.

This PR adds a monstrosity, a kinetics model that is a falloff and a three body process. Fondamentaly what it does is computing the falloff:
equation
using the three-body description of M:
equation

Thus a new ```FalloffThreeBodyReaction``` object that contains everything the ```FalloffReaction``` has and the ```ThreeBodyReaction```.

A slight update of ```Reaction``` and ```ThreeBodyReaction``` was done also.  The vector of coefficients were stored in the ```Reaction``` object, thus creating a ```std::vector<CoeffType>(n_species,1.L)``` for all the reactions.  Now this vector lives only in the appropriate objects, thus the methods ```void set_efficiency(const std::string & ,unsigned int, const CoeffType )``` and ```CoeffType efficiency(const unsigned int ) const``` have been adapted.  It saves memory, I don't see any reason why we should not do it, I hope I'm not missing something.  It also makes it easier for the computations.  A last change is in the falloff objects themselves, instead of passing down the full vector of molar mass (and thus needing the efficiencies for Troe), only the value of M is passed down.

### DIFF
--- a/src/kinetics/include/antioch/falloff_threebody_reaction.h
+++ b/src/kinetics/include/antioch/falloff_threebody_reaction.h
@@ -224,15 +224,18 @@ namespace Antioch
 //dk_dT = F * dkfwd_dT + kfwd * dF_dT
 //      = F * kfwd * (dk0_dT/k0 - dk0_dT/(kinf/[M] + k0) + k0 * dkinf_dT/(kinf * (kinf/[M] + k0) ) )
 //      + dF_dT * kfwd
-    dkfwd_dT = f * kfwd * (dk0_dT/k0 - dk0_dT/(kinf/M + k0) + dkinf_dT * k0/(kinf * (kinf/M + k0)))
+    StateType temp = (kinf/M + k0);
+    dkfwd_dT = f * kfwd * (dk0_dT/k0 - dk0_dT/temp + dkinf_dT * k0/(kinf * temp))
              + df_dT * kfwd;
 
     dkfwd_dX.resize(this->n_species(), kfwd);
+
+    StateType tmp = f * kfwd / (M + ant_pow(M,2) * k0/kinf);
 //dkfwd_dX = F * dkfwd_dX + kfwd * dF_dX
 //         = F * epsilon_i * kfwd / ([M] +  [M]^2 k0/kinf) + kfwd * dF_dX
     for(unsigned int ic = 0; ic < this->n_species(); ic++)
       {
-        dkfwd_dX[ic] = this->efficiency(ic) * ( f * kfwd / (M + ant_pow(M,2) * k0/kinf) + df_dX[ic] * kfwd );
+        dkfwd_dX[ic] = this->efficiency(ic) * ( tmp + df_dX[ic] * kfwd );
       }
 
     kfwd *= f; //finalize
--- a/src/kinetics/include/antioch/falloff_threebody_reaction.h
+++ b/src/kinetics/include/antioch/falloff_threebody_reaction.h
@@ -182,8 +182,10 @@ namespace Antioch
         M += this->efficiency(s) * molar_densities[s];
     }
 
-    return (*this->_forward_rate[0])(T) / (ant_pow(M,-1) + (*this->_forward_rate[0])(T) /(*this->_forward_rate[1])(T)) * 
-            _F(T,M,(*this->_forward_rate[0])(T),(*this->_forward_rate[1])(T));
+    const StateType k0   = (*this->_forward_rate[0])(T);
+    const StateType kinf = (*this->_forward_rate[1])(T);
+
+    return k0 / (ant_pow(M,-1) + k0 / kinf) * _F(T,M,k0,kinf);
   }
 
   template<typename CoeffType, typename FalloffType>
--- a/src/kinetics/include/antioch/falloff_threebody_reaction.h
+++ b/src/kinetics/include/antioch/falloff_threebody_reaction.h
@@ -1,0 +1,272 @@
+//-----------------------------------------------------------------------bl-
+//--------------------------------------------------------------------------
+//
+// Antioch - A Gas Dynamics Thermochemistry Library
+//
+// Copyright (C) 2014 Paul T. Bauman, Benjamin S. Kirk, Sylvain Plessis,
+//                    Roy H. Stonger
+// Copyright (C) 2013 The PECOS Development Team
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the Version 2.1 GNU Lesser General
+// Public License as published by the Free Software Foundation.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc. 51 Franklin Street, Fifth Floor,
+// Boston, MA  02110-1301  USA
+//
+//-----------------------------------------------------------------------el-
+
+#ifndef ANTIOCH_FALLOFF_THREE_BODY_REACTION_H
+#define ANTIOCH_FALLOFF_THREE_BODY_REACTION_H
+
+// Antioch
+#include "antioch/antioch_asserts.h"
+#include "antioch/cmath_shims.h"
+#include "antioch/reaction.h"
+#include "antioch/lindemann_falloff.h"
+#include "antioch/troe_falloff.h"
+
+//C++
+#include <string>
+#include <vector>
+#include <iostream>
+
+namespace Antioch
+{
+  /*!\class FalloffThreeBodyReaction
+  * Base class for falloff processes coupled with efficiencies
+  *
+  * This class encapsulates a bastard between falloff and three-body reaction.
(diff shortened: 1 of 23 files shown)


src/kinetics/include/antioch/falloff_threebody_reaction.h
```diff
@@ -0,0 +1,249 @@
...
+#include <string>
+#include <vector>
+#include <iostream>
+
+namespace Antioch
+{
+  /*!\class FalloffThreeBodyReaction
+  * Base class for falloff processes coupled with efficiencies
+  *
+  * This class encapsulates a bastard between falloff and three-body reaction.
```
pbauman: I understand you don't like this model, but let's clean this up a bit. Put something like "Note: @SylvainPlessis *strongly* disapproves of the use of this model on the grounds ... <fill in your reasoning here>."

src/kinetics/include/antioch/reaction_enum.h
```diff
@@ -40,7 +40,12 @@ namespace Antioch
                         DUPLICATE,
                         THREE_BODY, 
                         LINDEMANN_FALLOFF,
-                        TROE_FALLOFF};
+                        TROE_FALLOFF,
+// I'm crying inside... This is to support the
```
pbauman: Adding your note above will be enough. This comment is unnecessary.

pbauman:
GRINS `make check` works with this. Update according to my comments above, as well as any comments from @roystgnr, and I'm good with this.

src/kinetics/include/antioch/reaction_enum.h
SylvainPlessis: Can't I at least put a *--throw-physics-out-of-the-window*?

roystgnr:
I'm getting a failure with VexCL on this one.  Let me see what I can figure out.

src/kinetics/include/antioch/troe_falloff.h
```diff
@@ -242,7 +236,7 @@ namespace Antioch
   {
      
     // Fcent = (1.-alpha)*exp(-T/T***) + alpha * exp(-T/T*) + exp(-T**/T)
-    StateType Fc = (1. - _alpha) * ant_exp(-T/_T3) + _alpha * ant_exp(-T/_T1);
+    StateType Fc = (1.L - _alpha) * ant_exp(-T/_T3) + _alpha * ant_exp(-T/_T1);
```
roystgnr: Why did we switch from 1. to 1.L (which forces conversions to long double) rather than 1 (which would still be accurately treated as CoeffType and then StateType) here?

roystgnr:
Okay, there's extraneous long double all over the place in troe_falloff.h.

Usually, all you need is nothing.  The compiler will convert "1 - alpha" to whatever floating point type alpha is, even if you leave 1 an integer; you're done.

For non-integers, the safest thing to do is declare e.g. "CoeffType(0.4L)".  Even in that case, the L is only necessary if you have a decimal number (like 0.4) that can't be represented exactly as a double.

SylvainPlessis:
Ok so if I understand correctly:
integers are properly converted because they are exactly represented in base 2, real numbers requires long double type only if float and/or double can't have them exactly within their bits.

That's the idea?

SylvainPlessis:
Can you please verify that I got it right?  And there might be some more cleaning to do, mostly in the tests.

roystgnr:
Almost right.  Ideally we want everything to convert to CoeffType.  For integers that happens automatically, as long as the value hits a binary operation with another CoeffType before it hits a division-by-integer.  For literals that convert exactly in base 2, we still want e.g. "CoeffType(0.5)" so that we get a float rather than a double when CoeffType is float.  For literals that don't convert exactly in base 2, the best thing to do is e.g. "CoeffType(2)/CoeffType(5)", but until/unless some wiseguy decides to try out software-defined-precision arithmetic, "CoeffType(0.4L)" will work fine.

SylvainPlessis:
So I guess the best options for the lines of the type
```
StateType n = 0.75;
```
would be
```
StateType n = constant_clone(T,CoeffType(0.75L));
```
which would take care of both precision stuff (short of the software-defined-precision arithmetic wise guy check) and vectorization.

Is this correct or am I missing something?

roystgnr:
Assuming T is an input of StateType and n is going to need to be of StateType, that's correct.

SylvainPlessis:
Ok then I have a few things to modify.

SylvainPlessis:
Everything passes nice and well for me.

roystgnr:
I'm making a half-assed attempt at resolving those conflicts before testing; since there's no conflicts in the library itself I won't worry about it too much.

src/kinetics/include/antioch/troe_falloff.h
```diff
@@ -196,40 +196,34 @@ namespace Antioch
...
     // Note log10(x) = (1.0/log(10))*log(x)
-    StateType  c = - 0.4 - _c_coeff*logFcent;
+    StateType  c = - CoeffType(0.4L) - _c_coeff * logFcent;
 
     // n = 0.75 - 1.27 * log10(Fcent)
     // Note log10(x) = (1.0/log(10))*log(x)
-    ANTIOCH_AUTO(StateType) n = 0.75 - _n_coeff*logFcent;
-    ANTIOCH_AUTO(StateType) d = Antioch::constant_clone(T, 0.14);
+    ANTIOCH_AUTO(StateType) n = CoeffType(0.75L) - _n_coeff * logFcent;
+    ANTIOCH_AUTO(StateType) d = constant_clone(T,0.14L);
```
roystgnr: Need to wrap this 0.14L in a CoeffType() conversion.

Alternatively, I could add the downconversion in our utilities/ implementations.  I kind of like being alerted to downconversions, though, so we don't make them by accident.

src/kinetics/include/antioch/troe_falloff.h
```diff
@@ -309,25 +297,25 @@ namespace Antioch
     // Compute log(Fcent) once
     StateType logFcent = ant_log(Fcent);
     // n and c and derivatives
-    StateType  d = Antioch::constant_clone(T, 0.14);
-    StateType  c = - 0.4 - _c_coeff * logFcent;
-    StateType  n = 0.75 - _n_coeff * logFcent;
+    StateType  d = Antioch::constant_clone(T, 0.14L);
```
roystgnr: Likewise here

SylvainPlessis:
Let's hope this time they all are here.

roystgnr:
Can you rebase on master and fix up the conflicts?  I'm now seeing some regression tests failures. Looks like it might be just too-tight tolerances, in which case it'd be easier for me to send you a PR than for you to guess how many epsilons my system needs.

SylvainPlessis:
Alright rebasing get all the changes done during the last merge go away.  For some reason that eludes me, redoing these changes (adapting the new models to the ```KineticsConditions``` object) won't make the tests pass with the old tolerance (quite strong I admit), and we lose float in the process.  I'm merging against master to see if it changes anything.

If not, what version do you prefer, the rebased or the merged?

SylvainPlessis:
Merging gets the very same thing, float time is over for falloff models…

SylvainPlessis:
I choose rebasing, there is no big reason to choose one over the other, and the advantage is that the commit of adapting the models to the ```KineticsConditions``` object is now well defined and not lost in a merge.

roystgnr:
Change the weird "300*epsilon or 5e16" stuff to just 1000*epsilon?  That keeps the troe long double tests working for me but allows us to play with extended precision with more comfort later.

As for the float tests... I don't know *what* is happening there.  We're going an order of magnitude over tolerance in the troe tests, and I'd just ascribe that to "machine roundoff error isn't perfectly linear in epsilon".  If the math is rough enough to take away a couple orders of magnitude precision in the double case, I'd not be surprised that it's taking away even more from floats.

The Lindemann results are *nuts*, though.  Tolerance of 8e-6, diffs of 1e-2?  That suggests some kind of "small_num + big_num - big_num" cancellation going on under the hood.  The float results might be a canary in the coal mine here.

roystgnr:
Anyway, we're passing with all the apps and extras; nothing but the float tests giving us trouble.  I'm concerned but I'm still okay to merge.

SylvainPlessis:
I found the error with float: if you go to high enough temperature (~2000 K), the rate constants between ```k0``` and ```kinf``` are so different that the denominator ```1/M + k0/kinf``` is very inaccurate at the float precision.  Limiting the temperature range at 1500 K solves this issue.  We should make a note about this in the documentation.

roystgnr:
Nice!  Add that note and we're good to merge.  :+1:

SylvainPlessis:
Thanks for this remark, I have no idea why I didn't do it in the first place...

pbauman:
:+1:

SylvainPlessis:
Be aware that we should tag the version of Antioch just after this merge.
For the add part in the log, we add, physically speaking:
  - kinetics theory for diffusion, viscosity and thermal conduction
  - the three-body falloff kinetics model

Technically speaking, a hell lot of things, does anyone need the list?

pbauman:
I just opened a thread in antioch-devel; let's keep the release discussion to that thread, please.

PR #99, opened 2014-12-18
Parsing management

SylvainPlessis:
More parsing joy.

The idea here is to provide a more general ascii parser:
the data are formatted by columns, the added features enables to provide ignored columns.  See the input file for an example *test/input_files/test_parsing.ASCII*, where the first column (names are not counted) is ignored.  Thus providing to the parser the index of the column (zero here):
```
    Antioch::ASCIIParser<Scalar> parser(filename,true);

    parser.set_ignored_columns(std::vector<unsigned int>(1,0));
```

Thus overloads were necessary to implement this capability, this is simply the same methods (constructors and parsing methods) but having a parser in the input parameters instead of a filename and a boolean for verbosity.  It enables to parameterize the parser before using it.

This PR is to be applied anytime after PR #92 (transport_thermo_parse).

Only the very last commits (18th december) are to be reviewed here.


pbauman:
Do we merge this now or was it part of #117?

SylvainPlessis:
This one is not obsolete, though the capacity it adds requires much much less commits.  It has also the transport_new history, so let's consider this one obsolete and wait for the PR that will supersede it.  Right now, I think it's best if you concentrate on PR#101 and PR#115, those can be added independently, I'll take care of those three lasting PR, making them shorter in terms of commits and merging as properly as I can #98.

PR #102, opened 2015-02-05
Conditions reac

roystgnr:
This rebases #74, and adds one new commit full of shim methods to try and keep API compatibility.  Tests all pass with GRINS; I'll at least double-check FIN-S too before merging, and Suzerain too if (as I expect) that builds more nicely on our mkl-equipped machines than it did on my laptop.


roystgnr:
Travis failure looks like memory issues with the gcc-4.6 + VexCL combination, which (several GB of RAM later) works on two other systems I've tried.

FIN-S is happy with this branch.

pbauman:
:+1: when you're happy.

SylvainPlessis:
I'm all for it

roystgnr:
I'm still having trouble getting https://github.com/RhysU/suzerain to recognize my mkl module, so I'm going to pull as-is, and if @RhysU finds a problem later then he can yell at me for it.  :-)

Issue #20, opened 2013-09-16
Support reaction chemistry in S3D


nicholasmalaya:
We want to enhance antioch to fully support the reaction chemistry in S3D. Using Jackie Chen's paper on [S3d](http://iopscience.iop.org/1749-4699/2/1/015001) as a model document, we the following capabilities: 

* Mixture-averaged diffusion coef. determined from binary diffusion coef and the mixture composition. 
* Only treat (N-1) species, with the last coming from constraint (e.g. \sum Y_i = 1)

S3D assumes an ideal gas mixture, so the equation of state is all set. I need to read through the documentation more carefully, and then will propose Antioch files to modify.

SylvainPlessis:
Isn't this the whole transport PR craziness?

Shouldn't we delete this issue?

nicholasmalaya:
Yes. This issue is deprecated. Might be worth plopping a few references to the actual tickets that resolved this issue before closing it, however.

SylvainPlessis:
Done.

PR #103, opened 2015-02-05
Transport/kinetics theory

SylvainPlessis:
First PR changing PR #79.

It adds the files, and all the necessary arrangements for the models.  Changes are:
 - the tree structure has a ```src/diffusion```, a ```src/viscosity``` and a ```src/thermal_conduction``` folder(s?), models were moved appropriately
 - the kinetics theory models, obviously, were added, the dependencies are:
     - transport species to store required value (dipole moment, Lennard-Jones values, etc)
     - transport mixture to play the equivalent role of chemical mixture
     - Stockmayer potential for diffusion and viscosity
     - a spline, GSL here, for spline the Stockmayer potential
     - the rotational relaxation for the thermal conduction
     - addition of physical constants, Boltzman and others
     - addition of the Debye unit for the dipole moment
     - the *Cv_(vib,tr,rot)_over_R* methods in ```StatMechThermodynamics```

The tests are added also, all compiles well and nice on my computer, vectorized versions and all.

The commits are much more verbose and intelligible, much less also.  This PR is totally and absolutely backward compatible because the new things are not used yet.

Next PR will be about integrating the physical models into a ```PhysicalSet``` object, with all the metaprogramming around and the parsing/building.

Last one will be the actual implementation of those models in ```WilkeEvaluator```, this is where there will be the API breaks, if only because the diffusion will be included into the ```WilkeEvaluator```


SylvainPlessis:
And another change which is the travis configuration, this is where we suppress the vexcl tests and add gsl dependence.  Normally it should pass now.

SylvainPlessis:
rebased

roystgnr:
It'll need a couple conflicts manually patched up (and one *non*-conflict in .travis.yml - I'd like to leave the vexcl dependencies commented out, since even if we can't use it now we might be able to reenable it later) after we pull #104, but nothing difficult.

I haven't looked deeply enough at the code to be sure I'm happy with the new features, but I've looked well enough to be content that you're probably not breaking any old features.  ;-)

No, wait, spoke too soon.  With VexCL and MetaPhysicL enabled I'm seeing segfaults in rotational_relaxation_vec_unit and kinetics_theory_thermal_conductivity_vec_unit - I'll see if I can figure that out.

roystgnr:
Changing from cast-to-float to cast-to-CoeffType in rotational_relaxation.h:65 fixes the core dumps for me, so our internal tests are working.

Testing with GRINS and FIN-S now.

roystgnr:
Strictly speaking I'm testing my own version of those last two commits, but since yours looks effectively identical I won't bother re-updating and repeating.  GRINS tests just passed.  Unless @pbauman screams, I'll merge as soon as the last FIN-S tests are done.

src/transport/include/antioch/Stockmayer_potential.h
```diff
@@ -0,0 +1,239 @@
...
+  template <typename CoeffType>
+  inline
+  StockmayerPotential<CoeffType>::~StockmayerPotential()
+  {
+     return;
+  }
+
+  template <typename CoeffType>
+  inline
+  void StockmayerPotential<CoeffType>::init()
```
pbauman: So StockmayerPotential is not just a function, but it is also number? I.e. the flood of hardcoded data is always tied to this and  StockmayerPotential will never mean anything else?

src/utilities/include/antioch/gsl_spliner.h
```diff
@@ -0,0 +1,179 @@
...
+
+// GSL
+#include <gsl/gsl_spline.h>
+#include <gsl/gsl_errno.h>
+#include <vector>
+
+namespace Antioch
+{
+
+  template<bool B>
```
pbauman: I'm not grokking what's going on here.

roystgnr:
Gah, should have reloaded the page before hitting merge...

pbauman:
It's no problem. If changes are needed we can deal with it,

src/utilities/include/antioch/gsl_spliner.h
roystgnr: The meta-level is easy to understand: @SylvainPlessis doesn't comment enough, by an order of magnitude.  ;-)

SylvainPlessis:
This adds the physical models necessary to issue #20.

PR #105, opened 2015-02-06
Transport/kinetics theory physical set

SylvainPlessis:
The second PR to replace #79.
Finally there will be four instead of three, I've sliced this one in two parts: the implementation of the metaprogramming stuff and the parsing corresponding to the new design.  
This is the metaprogramming stuff.  Note that all tests pass and this is still fully backward compatible because no new things is still used in at the Wilke level, and the missing files called in the *_utils.h* files are not called for the simple reason that no one include the *_utils.h* files yet.

So basically we have a wrapper around all the transport physics, ```PhysicalSet```, that deals with all the different physics and notably the common part of creation, adding a model and deleting.

The design is declared in *physics_metaprogramming_decl.h*, the default are implemented in *physics_metaprogramming.h*.  Then all the specializations are declared in *model_utils_decl.h* and implemented in *model_utils.h*, to follow the very same structure as the vectorization.


SylvainPlessis:
rebased

SylvainPlessis:
rebased

roystgnr:
I merged instead of rebased, but there were no conflicts and the result (naturally) passed with GRINS for me.  I'm not sure this is exactly the final design we want to use, but it looks decent and IMHO ready to merge once my FIN-S tests pass and @pbauman gives it a look.

roystgnr:
FIN-S tests pass.

pbauman:
Agree that we'll want to look more closely at the design in the future (in Sylvain's defense, this design problem is *hard*), but it definitely has much more capability and is probably in the right direction. Also looks like he addressed my critiques from the telecon a week or two ago. :+1:

SylvainPlessis:
This adds the usability of necessary models for issue #20.

PR #111, opened 2015-02-06
Species parsing

pbauman:
Replaces #88. Just merged from master (only one conflict due to whitespace).


src/core/include/antioch/chemical_mixture.h
```diff
@@ -73,12 +73,14 @@ namespace Antioch
   class ChemicalMixture
   {
   public:
+    template <typename Parser = ASCIIParser<CoeffType> >
```
pbauman: Why template around the parser type?

SylvainPlessis:
Because you can use an xml file, or a chemkin file, or in the future a hdf5 file, and the constructor has to know it to instantiate the correct parser.

SylvainPlessis:
Well at least that's what is wanted, being able to read from any format implicitly

roystgnr:
Yeah, but this should probably be a runtime decision (i.e. with virtual functions if possible or switch statements if not), not a compile-time decision (with templates).  We use templates for state types because we don't want a virtual function call every time we add two numbers; but we *can* afford a virtual function call every time we instantiate a parser.

SylvainPlessis:
The thing is ```ChemicalMixture``` requires the mixture to be known at construction time, with the only information being file names.  All being implicit, it was easier to template around.

roystgnr:
Templating isn't about "do you decide at construction time or at use time"; almost the exact opposite!  Templates vs virtual functions are about "do you decide at compile time or run time".  The cost of deciding at compile time is that the size of your compiled code is exponential in the number of template combinations.  The cost of deciding at run time is that you need to make a virtual function call for each decision.  If the decisions aren't in inner loops (e.g. if you're making the decision at construction time), the virtual function call option is greatly preferable.

SylvainPlessis:
I see…  I have to admit, I kinda like templates, I may have been a little carried away…
So a base ```Parser``` class and derived objects I guess.

pbauman:
> So a base `Parser` class and derived objects I guess.

:+1:

pbauman:
> I see… I have to admit, I kinda like templates, I may have been a little carried away…

Learning how new toys work is fun, but have to resist the temptation to use it all the time. :)

pbauman:
BTW, @SylvainPlessis, if you want to branch from this branch, merge from master, and open a new PR with those changes, go for it.

pbauman:
Just please be sure to reference this in the new PR.

SylvainPlessis:
I'll check first if it's not easier to use the original branch, after all, we're just a merge away from it.

pbauman:
We went back to #88. Closing this.

PR #113, opened 2015-02-07
Xml header fix

pbauman:
This header got removed in #72 and we need it for backward compatibility. Note I've used the `#warning` is not standard, but seems to be accepted, so far. We'll see what clang says. Would also like to see what Intel says. I'm open to other suggestions, but at least wanted a warning printed at compile time to clue the user in as to what the not-deprecated header is.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -58,6 +58,7 @@ pkginclude_HEADERS += kinetics/include/antioch/kinetics_evaluator.h
 pkginclude_HEADERS += parsing/include/antioch/tinyxml2.h
 pkginclude_HEADERS += parsing/include/antioch/tinyxml2_imp.h
 pkginclude_HEADERS += parsing/include/antioch/read_reaction_set_data.h
+pkginclude_HEADERS += parsing/include/antioch/read_reaction_set_data_xml.h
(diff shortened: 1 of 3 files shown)


roystgnr:
I could swear I *did* exactly this already, but git sure doesn't seem to think so.  Looks like the reason I never noticed the regression was that "make install" doesn't remove old installations' files and I've been using the same old install prefix for testing new versions.  That's embarrassing.

I'll start testing this, but if you want to merge then don't wait for me.  "Might be broken" is still a huge improvement over "definitely broken".

pbauman:
> I could swear I did exactly this already, but git sure doesn't seem to think so.

Some of these rebases were scary (which is why I was just merging) - wouldn't surprise me if it got lost in a rebase.

> Looks like the reason I never noticed the regression was that "make install" doesn't remove old installations' files and I've been using the same old install prefix for testing new versions. That's embarrassing.

Sadly, I did the same thing for a bit.

> I'll start testing this, but if you want to merge then don't wait for me. "Might be broken" is still a huge improvement over "definitely broken".

OK, in that spirit, I'll merge this and we'll cross the CPP macro bridge if we come to it.

Issue #110, opened 2015-02-06
Fix GSL/build system


pbauman:
1. GSL should be optional - I got a configure error when trying to configure without GSL.
2. Adding GSL, make check is failing to build the spliner test correctly. Probably not passing all the flags correctly.

Will get to this after AnitochCon2015 is over with...

SylvainPlessis:
--without-gsl works as expected on my system

SylvainPlessis:
By the way, should we indeed not pass the tests if GSL is not here or should we make them skipped tests?
I'd prefer the latest version.

pbauman:
> --without-gsl works as expected on my system

Damn, really? That didn't configure for me. I'll play with this tonight.

roystgnr:
Configured and passed make check for me.  Which has me pretty damn suspicious; there should have been skipped or failed checks.  I may try actually uninstalling GSL and see what happens.

roystgnr:
Ha, never mind.  Didn't notice that I'd passed fewer tests than before.  The tests are disabled, not just skipped.

pbauman:
I think I have a handle on what's going on and will fix later tonight. I think y'all have gsl-config in your path, even without the module (i.e. /usr/bin/gsl-config). I don't so the AC_MSG_ERROR gets tripped even though it's not a required package. Also, I'll update the tests to show SKIPPED when GSL is not installed.

SylvainPlessis:
Do it also for the *stat_mech_thermo_unit_eigen*, I just followed this example and it's not very satisfying when actually thinking about it…

SylvainPlessis:
Is this issue still relevant?

pbauman:
yes. Haven't been able to get the time to fix it yet. Please leave open.

PR #114, opened 2015-02-07
Display skipped tests

pbauman:
Announce when GSL and Eigen tests are being skipped. Also updated config_summary.m4 to show GSL linking status. Also the CPPFLAGS fix should take care of 1 of my complaints in #110.


pbauman:
This is a fairly benign patch, so I'm going to merge.

PR #116, opened 2015-03-02
Transport/thermo

SylvainPlessis:
This supersedes PR #91.

graph

It's built on the new PRs instead of the old ones, it does not add a lot, only simplifies the ```WilkeTransportMixture``` object by supressing the ```Thermo``` templates, as a ```ThermoHandler``` is in the ```TransportMixture```.

The others changes are for backward compatibility (the ```WilkeMixture``` object for instance is not more derived from the new one, I couldn't find an elegant way of doing it), the building methods are adapted so only the micro thermo is passed to the physics needing it (Euler dffusion and kinetics theory thermal conduction).

Tests are updated to use latest objects.


roystgnr:
Works with GRINS+FIN-S+MetaPhysicL+VexCL for me.

src/core/include/antioch/chemical_mixture.h
```diff
@@ -569,6 +585,24 @@ namespace Antioch
...
+     antioch_deprecated();
+     return *this;
+  }
+
+  template<typename CoeffType>
+  template<typename ThermoEvaluator>
+  inline
+  const ThermoEvaluator & ChemicalMixture<CoeffType>::thermo_evaluator() const
+  {
+     std::cerr << "What the hell are you doing here? This backward compatibility place should not be explored" << std::endl;
```
pbauman: I see you left a nice message for me. :)

pbauman:
Cool. I don't completely grok everything here yet, but I don't see anything egregious. Barring any complaints from @roystgnr, this is good to merge.

src/core/include/antioch/chemical_mixture.h
roystgnr: Not a nice message.  Output deep in the library needs to be wrapped in antioch_do_once() (as soon as we port over libmesh_do_once) so it can't end up being spammy.

In this case, though, the output should just be "antioch_deprecated();"

roystgnr:
I'd say fix up the antioch_deprecated() and then hit the button.

src/core/include/antioch/chemical_mixture.h
SylvainPlessis: Well that's a place that exists strictly so the compiler can compile. I was not sure what to do here, the only thing I'm convinced is that this piece of code should be short-lived.

roystgnr: Hmm... github didn't show me the "antioch_error();" below it.  That fixes the spamminess problem, but just raises further questions:

This is a backwards compatibility shim that's been created to fail at run time rather than compile time?  That's not an improvement...

roystgnr:
The new comment is better, and Sylvain's explanation in person was better still.  There's no way either old or new code should be hitting that function; it's just an artifact of the refactoring.

PR #117, opened 2015-03-02
Transport/parser

SylvainPlessis:
Supersedes PR #92.

Simply updates the transport species parsing.

Now it's a big overload as templated methods can't be virtual, so as we have a ```BaseParser``` which contains all the methods.  In the derived object, only ```ASCIIParser``` currently, the six methods (or so) are all pointing to a templated root method.

Updated graph to follow.


roystgnr:
Diffs look good to me.  Tests are all happy.  :+1:

pbauman:
Sweet.

PR #118, opened 2015-03-04
Opt vis

SylvainPlessis:
Supersedes PR #101, this one being actually branched off master instead of wilke_approx.

Small and self explanatory I think.


pbauman:
Have you timed the speed up yet, just for giggles?

SylvainPlessis:
5 times faster for the Exact computations, full application scale (transport + kin).

SylvainPlessis:
Na, 20% speedup, there is a factor 5, but not quite at this place….

roystgnr:
Looks good.  :+1:  Do we have any application regression test coverage for the new algorithm, or just the antioch tests?

SylvainPlessis:
We can use the ExacT code.  Last time I've checked, we were still agreeing.

pbauman:
Awesome!

PR #119, opened 2015-03-05
Transport/opt

SylvainPlessis:
supersedes PR #93.

graph

Changes are made as explicit as possible with commit messages.  Only ```WilkeTranportEvaluator``` and ```PhysicalSet``` takes ```KineticsConditions```.

### DIFF
--- a/src/core/include/antioch/kinetics_conditions.h
+++ b/src/core/include/antioch/kinetics_conditions.h
@@ -29,6 +29,7 @@
 // Antioch
 #include "antioch/antioch_asserts.h"
 #include "antioch/particle_flux.h"
+#include "antioch/temp_cache.h"
 
 // C++
 #include <vector>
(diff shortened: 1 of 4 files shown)


src/core/include/antioch/kinetics_conditions.h
```diff
@@ -109,7 +115,14 @@ namespace Antioch{
...
   {
-     return _temperature;
+     return _temperature.T;
+  }
+
+  template <typename StateType, typename VectorStateType>
+  inline
+  const StateType & KineticsConditions<StateType,VectorStateType>::Tvib() const
+  {
+     return _temperature.T;
```
pbauman: Maybe I'm missing something, but it seems like this method is for vibrational temperature and we're returning translational temperature.

SylvainPlessis: Yep, and so far, all we have is Tv = T.

SylvainPlessis:
Just ping me when we want this one merged, the conflicts were expected and should be on the method's name (``` StateType & T``` v/s ```KineticsConditions<> & conditions```).

I'll merge master in it at the last moment, to be sure it requires only one merge.

roystgnr:
Putting it through the tests at the moment.

src/core/include/antioch/kinetics_conditions.h
pbauman: Let's at least put a comment in to that effect so we don't inadvertently deceive anyone.

pbauman:
Barring adding the comment, I'm good with this if Roy's tests pass.

roystgnr:
All good; ready to merge.

SylvainPlessis:
On it.

PR #120, opened 2015-03-05
Kinetics/opt

SylvainPlessis:
Supersedes PR #96.

graph

It overloads the rate constant calculations for all the models having a ```T^\beta * exp(-Ea/T)``` part with a ```KineticsConditions``` objects to compute instead ```exp(\beta * lnT - Ea/T```.

### DIFF
--- a/src/core/include/antioch/kinetics_conditions.h
+++ b/src/core/include/antioch/kinetics_conditions.h
@@ -29,6 +29,7 @@
 // Antioch
 #include "antioch/antioch_asserts.h"
 #include "antioch/particle_flux.h"
+#include "antioch/temp_cache.h"
 
 // C++
 #include <vector>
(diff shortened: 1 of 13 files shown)


roystgnr:
Nice little optimization.  Does it get you much in the benchmarks?

SylvainPlessis:
It was felt, I can't remember how much, but it did a non trivial difference.

SylvainPlessis:
And it just occured to me we should do the same with ```1/T```

roystgnr:
Caching 1/T might be a good idea.  Leave this branch unchanged, though?  We should be good to merge it shortly.

SylvainPlessis:
I absolutely don't intend to touch anything to the current PRs, short of conflict solve.

roystgnr:
MetaPhysicL+VexCL+GRINS+FIN-S are happy.  Diffs look good to me.  :+1: whenever @pbauman is happy too.

Issue #122, opened 2015-03-05
Parser not accepting units="K" for activation energy


pbauman:
I get the follow error for the xml file that I'll post here when I try and use Antioch kinetics.

```
/fry1/data/shared/software/libs/antioch/master/gcc/4.8.4/gsl/1.16/include/antioch/read_reaction_set_data.h, line 188, compiled Mar  5 2015 at 09:14:13
terminate called after throwing an instance of 'Antioch::UnitError'
  what():  Error in reaction O + O [=] O2
Ea parameter's unit should be homogeneous to "J/mol", or "K" and you provided "K"
Aborted (core dumped)
```

pbauman:
Here's the XML file I wanted to use
```
<?xml version="1.0"?>

<ctml>
  <validate reactions="yes"/>

  <!-- phase bunsen     -->
  <phase dim="1" id="ozone">
    <speciesArray>O O2 O3</speciesArray>
  </phase>

  <reactionData id="LinanRodrigues1985">
    <!-- reaction 0001    -->
    <reaction reversible="no" type="threeBody" id="0001">
      <equation>O + O [=] O2</equation>
      <rateCoeff>
        <Arrhenius>
           <A units="cm6/mol2/s">4.12e+14</A>
           <E units="K">-767</E>
        </Arrhenius>
        <efficiencies defaults="1.0"></efficiencies>
      </rateCoeff>
      <reactants>O:2.0</reactants>
      <products>O2:1.0</products>
    </reaction>

    <!-- reaction 0002    -->
    <reaction reversible="no" type="threeBody" id="0002">
      <equation>O2 + O [=] O3</equation>
      <rateCoeff>
        <Arrhenius>
           <A units="cm6/mol2/s">2.64e+13</A>
           <E units="K">-300</E>
        </Arrhenius>
        <efficiencies O3="0.92" O2="0.94" O="1.13"></efficiencies>
      </rateCoeff>
      <reactants>O2:1.0 O:1.0</reactants>
      <products>O3:1.0</products>
    </reaction>

    <!-- reaction 0003    -->
    <reaction reversible="no" type="threeBody" id="0003">
      <equation> O3 [=] O2 + O </equation>
      <rateCoeff>
        <Arrhenius>
           <A units="cm6/mol2/s">2.03e+15</A>
           <E units="K">12000</E>
        </Arrhenius>
        <efficiencies defaults="1.0"></efficiencies>
      </rateCoeff>
[...]

PR #123, opened 2015-03-05
Bugfix.

SylvainPlessis:
The unit was tested against the default unit, and this default
  unit is only for message purpose...

The *default_unit* is only to print the desired type of unit is none is provided, and I used it to compare...

Well it fixes the bug of issue #122.

### DIFF
--- a/src/parsing/include/antioch/read_reaction_set_data.h
+++ b/src/parsing/include/antioch/read_reaction_set_data.h
@@ -173,7 +173,7 @@ namespace Antioch
         bool found(false);
         for(unsigned int i = 0; i < accepted_unit.size(); i++)
         {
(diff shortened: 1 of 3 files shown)


pbauman:
Sorry to be a pain, but it's probably a good idea to update one of the xml files to have one reaction use this feature so we cover it.

pbauman:
In the regression suite that is.

SylvainPlessis:
Indeed, I was being lazy...

Compiling as of now.

pbauman:
This is good to merge. When whenever is good for you - I don't want to accidentally stomp on the other PRs open at the moment.

pbauman:
And thanks for the quick fix!

SylvainPlessis:
I've been bragging a little bit too much about unit management to let that...

PR #127, opened 2015-03-12
Clear Warnings

pbauman:
Clearing out some warnings I was hitting with GRINS. This may not be all of them yet because I'm not instantiating the new capabilities. Still have a couple `-Woverloaded-virtual` warnings to clear out, but will wait on #126 before finishing up. At least didn't want folks to duplicate effort.

Issue #126, opened 2015-03-12
Can we make XMLParser::Troe() const?


pbauman:
The ChemKin version is const (out of necessity I see) leading to warnings that the base wasn't. However, as the XML version is coded right now, it can't be const. @SylvainPlessis, anyway to make this const on the top of your head before I dig through the code? This was the last warning for me to clear up.

pbauman:
Err, the base version was const, only the XMLParser wasn't.

```
warning: ‘bool Antioch::ParserBase<NumericType>::Troe() const [with NumericType = double]’ was hidden [-Woverloaded-virtual]
          virtual bool Troe() const  {not_implemented(); return false;}

[snip]

 warning:   by ‘bool Antioch::XMLParser<NumericType>::Troe() [with NumericType = double]’ [-Woverloaded-virtual]
   bool XMLParser<NumericType>::Troe()

```

SylvainPlessis:
Alright so the situation here:
  - ```XMLParser``` during the Troe() operation sets the pointer *_Troe* to the correct value, so the method cannot be _const_.
  - ```ChemKinParser``` is caching all the information at the *bool reaction()* method level, therefore *bool Troe() const* is but a retrieving of the info and can be _const_.  As I _const_ methods as much as I can. I just went for it.
  - ```ParserBase```do not have a _bool Troe() const_ method…  What is the compiler trying to say?

I forgot this issue, the answer is to do with ```XMLParser``` as ```ChemKinParser``` parser does, which means retrieving all the info at the very beginning in the *bool reaction()* method.  For historical reason, the xml was written before chemkin and it was just easier to read along non _const_ than to process everything at one place, it is not the case.

If someone wants to go for it, good (and good luck), if not, I'll start on it next week.

pbauman:
> XMLParser during the Troe() operation sets the pointer _Troe to the correct value, so the method cannot be const.

Sorry for not being clear, but this was the crux of my question: can we modify the parsing algorithm to figure that if Troe should be true/false and make Troe() a const accessor?

SylvainPlessis:
> Sorry for not being clear, but this was the crux of my question: can we modify the parsing algorithm to figure that if Troe should be true/false and make Troe() a const accessor?

Yep it's possible, and it will be done.  The principle is to set all the reaction in the method _bool reaction()_.  *_Troe* for instance will be *NULL* or pointing to the correct xml child.  Then 
_bool Troe()_ will become
```
[...]

PR #128, opened 2015-03-12
Clear Warnings

pbauman:
Clearing out some warnings I was hitting with GRINS. This may not be all of them yet because I'm not instantiating the new capabilities. Still have a couple -Woverloaded-virtual warnings to clear out, but will wait on #126 before finishing up. At least didn't want folks to duplicate effort.

### DIFF
--- a/src/core/include/antioch/chemical_mixture.h
+++ b/src/core/include/antioch/chemical_mixture.h
@@ -362,7 +362,7 @@ namespace Antioch
   template<typename CoeffType>
   inline
   void ChemicalMixture<CoeffType>::read_species_characteristics(ParserBase<CoeffType> * parser,
-                                                                const std::string & species_data,
+                                                                const std::string & /*species_data*/,
                                                                 const std::string & vibration_data,
                                                                 const std::string & electronic_data)
   {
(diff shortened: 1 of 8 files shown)


SylvainPlessis:
While you're at it, there is one that PINC hits, in the file _src/kinetics/.../constant_rate.h_, line 155, the parameter T is unused.

Thanks for doing this, I've been meaning to do it quite a long time ago...

SylvainPlessis:
In _src/kinetics/include/antioch/reaction.h_, the methods ```compute_rate_of_progress_and_derivative``` (declared l. 265/277, defined l. 989/1156) contains a ```ChemicalMixture``` in their parameters' list.  This chemical mixture is quite old, it's odd we never hit it before, and is absolutely useless, both in the deprecated and new version.

We should at some point purely and simply delete it, thus modifying l. 285 of file _src/.../reaction_set.h_ where it is called.

Then there is a little bit more of hidden parameters is the parsers, I'll update the corresponding PR #131. 

Once this is done, PINC is alright with warnings from Antioch and I'm fully ok for merging.

roystgnr:
Sylvain pointed out that we haven't merged this to master yet.  Paul, why'd you close #127 ?  Don't we want these fixes in both branches?

pbauman:
Because we can manually merge to master. Will do so now.

PR #130, opened 2015-03-23
Better use of _mixture, now it's in charge of all the mixture related

SylvainPlessis:
I don't really know where to put that one, but since I'm considering one change to be a bugfix, I'll put it here.  How do we make master aware of this?

The change is in the ```WilkeTransportEvaluator```, there was two places (l. 375 & 382) where we recovered some ```ChemicalMixture``` method from the mixture associated to the viscosity.  The problem is that ```TransportMixture``` do not have those method (enough method duplication), thus this might end up as a compile-time error if you want a ```TransportMixture``` with your viscosity.

So now all the mixture-related things are managed by the ```WilkeTransportMixture```, which ended up in a few simplifications with the bugfix.

### DIFF
--- a/src/transport/include/antioch/wilke_transport_evaluator.h
+++ b/src/transport/include/antioch/wilke_transport_evaluator.h
@@ -168,7 +168,7 @@ namespace Antioch
      typename rebind<VectorStateType,VectorStateType>::type Ds(ds.size());
      init_constant(Ds,ds);
      _diffusion.compute_binary_diffusion_matrix(transport_conditions,cTot,Ds );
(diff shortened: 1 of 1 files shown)


SylvainPlessis:
Can we get this one to be merged in a near future? I just need the thumbs up and I'll do it.

pbauman:
:+1:

SylvainPlessis:
One thumb says I can do it tonight :waxing_crescent_moon: (yeah, I can emoji too :bowtie: )

roystgnr:
Sorry!  I finally glanced over the diffs last week but didn't find time to run tests.  I'll do that now and merge right away if everything is happy.

SylvainPlessis:
So now I just merge 0.3.0-release into master locally then push master?  I'm not sure about the protocol in use there.

pbauman:
Yea, I'll take care of it now. But I manually merge 0.3.0-release into master and push to master.

PR #131, opened 2015-03-23
Parser/const

SylvainPlessis:
```XMLParser::Troe()``` and ```ParserBase::Troe()``` are ```const```'ified.

It finally required little changes as the only thing that really needed to be stored was the Troe environment pointer.

Also added an id to the virtual methods of the base object to know which one is asked for and not available.

### DIFF
--- a/src/parsing/include/antioch/parser_base.h
+++ b/src/parsing/include/antioch/parser_base.h
@@ -117,16 +117,16 @@ namespace Antioch
 
 /// species
         //! reads the species set
-        virtual const std::vector<std::string> species_list() {not_implemented(); return std::vector<std::string>();}
+        virtual const std::vector<std::string> species_list() {not_implemented(0); return std::vector<std::string>();}
 
         //! reads the mandatory data, not valid in xml && chemkin
-        virtual void read_chemical_species(ChemicalMixture<NumericType> & /*chem_mixture*/)  {not_implemented();}
+        virtual void read_chemical_species(ChemicalMixture<NumericType> & /*chem_mixture*/)  {not_implemented(1);}
 
         //! reads the vibrational data, not valid in xml && chemkin
-        virtual void read_vibrational_data(ChemicalMixture<NumericType> & /*chem_mixture*/)  {not_implemented();}
+        virtual void read_vibrational_data(ChemicalMixture<NumericType> & /*chem_mixture*/)  {not_implemented(2);}
 
         //! reads the electronic data, not valid in xml && chemkin
-        virtual void read_electronic_data(ChemicalMixture<NumericType> & /*chem_mixture*/)  {not_implemented();}
+        virtual void read_electronic_data(ChemicalMixture<NumericType> & /*chem_mixture*/)  {not_implemented(3);}
 
 // transport, the thermo is explicit...
 
@@ -135,21 +135,21 @@ namespace Antioch
         virtual void read_transport_data(TransportMixture< ThermoHandler < NumericType, 
                                                                            NASAEvaluator<NumericType,NASA7CurveFit<NumericType> >,
                                                                            StatMechThermodynamics<NumericType> >,
-                                                           NumericType > & /*transport_mixture*/)  {not_implemented();}
+                                                           NumericType > & /*transport_mixture*/)  {not_implemented(4);}
 
         //! reads the transport data, not valid in xml && chemkin
         //  NASA9 + StatMech
         virtual void read_transport_data(TransportMixture< ThermoHandler < NumericType, 
                                                                            NASAEvaluator<NumericType,NASA9CurveFit<NumericType> >,
                                                                            StatMechThermodynamics<NumericType> >,
-                                                           NumericType > & /*transport_mixture*/)  {not_implemented();}
+                                                           NumericType > & /*transport_mixture*/)  {not_implemented(5);}
 
         //! reads the transport data, not valid in xml && chemkin
         //  CEA + StatMech for backward compat
         virtual void read_transport_data(TransportMixture< ThermoHandler < NumericType, 
                                                                            CEAEvaluator<NumericType>,
                                                                            StatMechThermodynamics<NumericType> >,
-                                                           NumericType > & /*transport_mixture*/)  {not_implemented();}
+                                                           NumericType > & /*transport_mixture*/)  {not_implemented(6);}
 
         //! reads the transport data, not valid in xml && chemkin
         //  NASA7 + Ideal Gas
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by SylvainPlessis** [comment]
Forgot to link it to issue #126

**Comment by SylvainPlessis** [comment]
The thermo reading methods hide each other, even if they're not supposed to be the same.

A little explanation.  For the thermo is a two level stuff, the mixture and the curve fit.  We have two curve fit options, ```NASA7``` and ```NASA9```.  Before this only ```NASA9``` was here under the name ```CEA```.  Thus the following deprecation/derivation pattern:

deprecated | derived from
--------------- | -----------------
```CEAThermoMixture<NumericType>``` | ```NASAThermoMixture<NumericType,CEACurveFit>```
```CEACurveFit<NumericType>```            |  ```NASA9CurveFit<NumericType>```

The new layout is the ```NASAThermoMixture``` with the choice of the curve, and apparently, the fact that ```CEAThermoMixture``` derives from ```NASAThermoMixture<,CEACurveFit>``` and ```CEACurveFit``` from ```NASA9CurveFit``` makes it that the four possibilities it offers for the thermo declaration hide each other if they're not explicit.

So corrected.

**Comment by SylvainPlessis** [comment]
There is some cleaning to be done…

Alright the previous message is not accurate, there is another deprecated object in the landscape: ```CEAThermodynamics``` which contains within itself the curve fits.

So we require a little bit more for compatibility purposes.

Now who uses this object?  We really should delete it.

Update on PR close to come.

**Comment by SylvainPlessis** [comment]
More about this ```CEAThermodynamics```…

When parsing for the transport properties, we use a ```TransportMixture``` that requires
 a ```ThermoHandler```, containing the macro thermodynamics.  Thus same thing than for the macro thermo alone, all methods have to be explicit as we cannot have virtual templated methods.

So parsing the transport properties requires to know the macro thermo evaluator, which is a ```NASAEvaluator<NumericType,CurveType>```, and for backward compatibility, a ```CEAEvaluator<NumericType>```.  

Funny thing, a ```CEAThermodynamics``` works like either a thermo mixture or a thermo evaluator, it can mimic both.
That's where it becomes hilarious: if ever for some dark reasons that will irreversibly stain your soul you want to use a ```CEAThermodynamics``` object and do some transport new style, you'll have trouble parsing stuff.  The parsers have not that possibility, and your compiler will refuse adamantly to compile your code (or you have stained its essence also and nothing can save your computer from eternal damnation).

I propose a simple and clear deleting of the culprit (both object and person if any) in a very near future.

**Comment by SylvainPlessis** [comment]
What about this one too?  I'd like to get clear on the release branch the soonest possible.

File: src/parsing/include/antioch/parser_base.h [hunk 1]
```diff
@@ -343,14 +346,15 @@ namespace Antioch
...
-  void ParserBase<NumericType>::not_implemented() const
+  void ParserBase<NumericType>::not_implemented(unsigned int i) const
   {
       std::cerr << "\n*********************************************************\n"
                 << "This method is not available with a " << _type << " parser.\n"
                 << "Parsing file " << _file << ".\n"
                 << "No format has been defined yet.  Maybe contribute?\n"
                 << "https://github.com/libantioch/antioch\n" 
-                << "\n*********************************************************\n"
+                << "\nMethod code: " << i
```
roystgnr: Could we switch all these to antioch_not_implemented()?  The error message here is nicer but we could move the generic parts of it to antioch_asserts.h and/or add an antioch_not_implemented_msg() option for non-generic additions.  The "Method code" here is *not* nicer than just getting a filename and line number in the output.

src/parsing/include/antioch/parser_base.h
```diff
@@ -173,100 +173,103 @@ namespace Antioch
...
-         virtual bool Troe()  {not_implemented(); return false;}
+         virtual bool Troe() const {not_implemented(16); return false;}
 
          /*! \return reaction id, 0 if not provided*/
-         virtual const std::string reaction_id() const  {not_implemented(); return std::string();}
+         virtual const std::string reaction_id() const  {not_implemented(17); return std::string();}
 
          /*! \return reaction equation */
-         virtual const std::string reaction_equation() const {not_implemented(); return std::string();}
+         virtual const std::string reaction_equation() const {not_implemented(18); return std::string();}
```
roystgnr: Shouldn't a lot of these simply be pure virtual functions?  It's much nicer to find out something is not implemented at compile time rather than run time.

SylvainPlessis:
Indeed, an antioch_not_implemented() should have been there, this code was me debugging, it worked nicely, I didn't go further than that, my bad.

Alright a quick talk and we have a protocol:
  - functions that are not ``universal'' (i.e. some formats simply do not exist), we keep it this way,
  - functions that are ``universal'' (everyone should parse this, and anyway, this is Antioch's mandatory) will be pure virtual.

SylvainPlessis:
Actually there's already this pure virtual thing, I did some stuff good :smile:

SylvainPlessis:
The details of the though process are in the comment of commit 222f97f, I reproduce it here.

In a very general way, only the change_file(const std::string & filename) and the
bool initialize() are critical.

The capacity to change file cannot be
specific to this or that parser type, because it is not parsing-related.
There is not format issue linked to it, we cannot afford to have a
parser that could be called to change a file and just throw an error.
The program would not notice because the old file is valid and
would be used.

The initialize for the kinetics is more of an historical legacy.  It is
used to avoid any useless operation if no kinetics information is
present in the file (l. 218 of read_reaction_set_data.h).  It is
easier for error-finding purpose to do it separately instead of
putting it in the bool reaction() method (l. 288), because this
will just consider that the reactions are read, and not that there is
no kinetics information in the file. Who knows what may come?  An easy
way out of it is to test the number of reactions before exiting,
maybe go this way?

Now every parser has different capabilities, and this is much easier
to consider that any call to any non-existing parsing capability is
equivalent to any other one, thus they all should be treated in the
same way, thus in the base object.

As of now, it's a warning, make it an error?

roystgnr:
I'd say just switch to antioch_not_implemented().  Then you get an error instead of a warning, and you get an automatically-updated filename and line number instead of an opaque manual code.

SylvainPlessis:
My bad (I think I need more coffee)… 
Actually antioch_not_implemented_msg() throws an antioch_not_implemented(), so this is a error and not a warning…
I did what you suggested, I added a antioch_error_msg(msg) and the not_implemented_msg() is the combination of this last one and the antioch_not_implemented.

I should have been verbose about that...

roystgnr:
I'm happy with the current version.  :+1: to merge tonight unless @pbauman has any concerns.

PR #132, opened 2015-03-25
EGlib correction to vanishing molar concentrations

SylvainPlessis:
This concerns the diffusion calculations, it computes modified diffusion terms to ensure computational stability with respect to vanishing concentrations.

This is the EG lib way of doing it, documentation of EG lib is [here](http://blanche.polytechnique.fr/www.eglib/).

I'd like some agreement on the choice of the value of epsilon before merging.

```
// EGlib traces management, see doc: http://blanche.polytechnique.fr/www.eglib/manual.ps
// page 5
// EGlib uses eps = 1e-16
       typename raw_value_type<StateType>::type eps(std::numeric_limits<StateType>::epsilon() * 10);
```


pbauman:
@SylvainPlessis did you ever notice any actual large sensitivity due to this issue?

SylvainPlessis:
@pbauman, actually I've never play chemistry in such conditions (almost pure).  They did seem concerned about that in the documentation.

PR #134, opened 2015-04-03
capability to remove a reaction

SylvainPlessis:
it decreases the index of all following reactions.

If for some reason you have a map of the reactions, this method requires to
update it, Antioch doesn't, so we're cool here.

This is the commit message, says pretty much all.

This method is needed for any purging of reactions.  For instance in PINC I want some system to be in stationary state, thus I need every molecule to have a chemical loss and a chemical production rate. This method enable to automatically modify by pruning the reaction set to avoid well species.


roystgnr:
Looks entirely reasonable.

PR #84, opened 2014-10-27
Kin reset

SylvainPlessis:
graph

Here is the branch for resetting the kinetics.  This is the first step, the type of resetting @pbauman talked about, the wrapper method with paranoid checks will come in a future PR.

##### Resetting
To reset your kinetics, you need first to get to your kinetics object, then you reset.  The tests (*kinetics_type*_unit.C) have been updated and provide nice examples on what to do once you have your rate object.
```
ReactionSet<Scalar> reac_set;
...do stuff...
// reinitialize reaction n
KineticsType<Scalar> & kinetics = reac_set.reaction(n).forward_rate();
VectorType new_values;
... fill new_values ...
kinetics.reset_coefs(new_values);
```
You can use at this level any method shown in the tests that uses the vector of values.  If you want to reset a single parameter (say Cf in a Kooij model), you need to change the ```KineticsType<Scalar>``` into the appropriate rate, here ```KooijRate<Scalar>```:
```
ReactionSet<Scalar> reac_set;
...do stuff...
// reinitialize reaction n
KooijRate<Scalar> * kinetics = static_cast<KooijRate<Scalar>*> (&reac_set.reaction(n).forward_rate());
Scalar new_Cf_value;
... define new_Cf_value ...
kinetics->set_Cf(new_Cf_value);
```

The subtleties about units are dealt with the parameters ```Tref``` and ```rscale``` (which is the value of R in the unit of given activation energy).  For details, see the doxygen documentation on the rate models (method ```reset_coefs```).

For third-body coefficients, the object ```Reaction<Scalar>``` already has what's necessary:
```
ReactionSet<Scalar> reac_set;
...do stuff...
// reinitialize third-body coefs of reaction n
Reaction<Scalar> & reac = reac_set.reaction(n);
... looping over species...
reac.set_efficiency(name_spec,id_spec,value_coef_spec);
```
And finally, for the falloff reactions, you can access the falloff model (Lindeman or Troe) using:
```
DesiredFalloffModel & falloff = reac_set.reaction(n).falloff()
... do whatever ...
```

##### For reviewers
Lots of lines, but always the same thing, mainly the ```reset_coefs``` method in the kinetics rates.  Only the tests for the reset of the kinetics objects have so far been made (update of the already existing tests), except for the photochemistry.  The falloff model fetch has not been added to the test (it's but a wrapper of the corresponding method in the ```FalloffReaction<Scalar>``` object).

PR #135, opened 2015-04-10
Resetting a specific parameter declared in input file

SylvainPlessis:
It'a work still in progress, but as feedback is really wanted, the whole thing will come to life before your eyes.  There's no [DO NOT MERGE] tag so here's a nice question one just to show there's more to it than meets the eye.

So the idea here is to be able to specify in an input file parameters we want to be perturbed, therefore reset.  Human-friendliness requires to have a string-to-parameter map, here's how it goes.

Below the kinetics part, the transport part should be easy to do already, but it's worth to go through it after the kinetics done, so we are comprehensive (same comment for thermo).

### context

A reaction has two levels, the chemical process and the kinetics model.  The XML parser already has string keywords for the parameters, ChemKin format is hard-coded so there's very little need of keywords (and no keyword for the parameters we're interested in), and ASCII can't parse the kinetics.  So XML keywords we go! They will be in italics for easy of read.
An example of every type of reaction can be found in the ```test/input_files/test_parsing.xml``` file.

A kinetics model contains a subset of the parameters *A, b, E, D, lambda, cross_section*. Now depending on the chemical process, we have different requirements as of how do we recognize them:
  - Elementary process: nothing special.
  - Duplicate process: in this we have k = sum_i ki(Ai, bi, Ei,...), so we need to know which one we're talking about.
  - Three-body: we need the *efficiencies* per species.
  - Falloff (whatever): it requires to know if we are in the low-pressure rate (*k0*) or the high-pressure rate (kinf - it's not a typo, there's no keyword).  I propose **0** and **inf**.
  - TroeFalloff: we need the parameters *alpha, T1, T2, T3*.
  - Three-body falloff: I don't want to talk about this...

### input file
So here's how the input file would show up, the reactions are taken from the ```test/input_files/test_parsing.xml``` file (we talked about this with Roy, but I still want to be sure you saw, read, agreed to it, please comment):
```
Antioch/0008/A/0
Antioch/0008/b/2

Antioch/0014/efficiencies/CH4

Antioch/0022/D/inf

Antioch/0032/E/0
Antioch/0032/alpha
```

So the reaction ```0008``` is a duplicate, the indexes range from 0 to 2.  We're asking here to modify the pre-exponential parameter of the first and the power parameter of the last, they're indexed in the order they come.  We should probably add the capability to explicitely give the order in the input.

Reaction ```0014``` is a three body, we're changing the efficiency of CH4 from the default of 1 to whatever.

Reaction ```0022``` is a Lindemann falloff, we're changing the Berthelot parameter of the high-pressure rate.

Reaction ```0032``` is a Troe falloff, we're changing the activation energy of the low pressure rate and the alpha parameter of the Troe model.

### the code

So the first step is to add an id to the reaction.  We already have those ids, the parsers provide them for warning messages purpose, so the first commit just change the ```Reaction``` object to store it and adds the setting line while parsing.

Next step will be to implement at the ```ReactionSet``` level the capacility to pinpoint a specific parameter and change its value.  A method of the taste:
```
void set_parameter_of_reaction(const std::string & reaction_id, const std::vector<std::string> & parameter_keywords, CoeffType new_value);
```
Fondamentally, this method would use a capability similar to the one described in PR #84, with the difference that you reset only one parameter and not the whole thing:
```
// find reaction_id, index r
... loop on reactions, test reaction_id == reaction(r).id() ...

... find the wanted parameter defined by the different keywords...

// either 
//   kinetics parameter this of rate number that
// or 
//   parameter this of chemical process
if(change_kinetics_model) // kinetics parameter
{
  reset_parameter_of_rate(reaction(r).forward_rate(that),this,new_values);
}else // chemical process parameter
{
  reset_parameter_of_chemical_process(this, new_value);
}
```

So we need to write the ```reset_parameter_of...``` methods and the details of the finding with appropriate checks.
The ```reset_parameter_of...``` methods are to be in the ```src/kinetics/.../kinetics_parsing.h``` and ```src/kinetics/.../reaction_parsing.h``` files.

Comments welcome.

### DIFF
--- a/src/kinetics/include/antioch/kinetics_enum.h
+++ b/src/kinetics/include/antioch/kinetics_enum.h
@@ -68,6 +68,16 @@ namespace Antioch
       return 1.0; // this HAS to stay this way because it is hard-coded for performances (see eq. above)
     }
 
+    enum Parameters{ A = 0,
+                     B,
+                     E,
+                     D,
+                     T_REF,
+                     R_SCALE,
+                     SIGMA,
+                     LAMBDA
+                   };
+
   } // end namespace KineticsModel
 
 } // end namespace Antioch
--- a/src/kinetics/include/antioch/kinetics_enum.h
+++ b/src/kinetics/include/antioch/kinetics_enum.h
@@ -76,7 +76,11 @@ namespace Antioch
                      T_REF,
                      R_SCALE,
                      SIGMA,
-                     LAMBDA
+                     LAMBDA,
+        // now for the falloff, we need to know if we want
+        // the low-pressure limit or HIGH
+                     LOW_PRESSURE,
+                     HIGH_PRESSURE
                    };
 
   } // end namespace KineticsModel
--- a/src/kinetics/include/antioch/kinetics_enum.h
+++ b/src/kinetics/include/antioch/kinetics_enum.h
@@ -68,7 +68,8 @@ namespace Antioch
       return 1.0; // this HAS to stay this way because it is hard-coded for performances (see eq. above)
     }
 
-    enum Parameters{ A = 0,
+    enum Parameters{ NOT_FOUND = 0,
+                     A,
                      B,
                      E,
                      D,
--- a/src/kinetics/include/antioch/kinetics_parsing.h
+++ b/src/kinetics/include/antioch/kinetics_parsing.h
@@ -55,7 +55,7 @@ namespace Antioch
   template <typename CoeffType, typename VectorCoeffType, typename ParamType>
   void reset_parameter_of_rate(KineticsType<CoeffType,VectorCoeffType> & rate,
                                KineticsModel::Parameters parameter,
-                               const ParamType & new_value);
+                               const ParamType new_value);
 
 
 //----------------------------------------
@@ -216,7 +216,7 @@ namespace Antioch
   template <typename CoeffType, typename VectorCoeffType, typename ParamType>
   void reset_parameter_of_rate(KineticsType<CoeffType,VectorCoeffType> & rate,
                                KineticsModel::Parameters parameter,
-                               const ParamType & new_value)
+                               const ParamType new_value)
   {
     switch(rate.type())
       {
--- a/src/kinetics/include/antioch/kinetics_parsing.h
+++ b/src/kinetics/include/antioch/kinetics_parsing.h
@@ -28,6 +28,7 @@
 
 //Antioch
 #include "antioch/antioch_asserts.h"
+#include "antioch/metaprogramming_decl.h"
 #include "antioch/kinetics_type.h"
 #include "antioch/constant_rate.h"
 #include "antioch/hercourtessen_rate.h"
@@ -55,7 +56,7 @@ namespace Antioch
   template <typename CoeffType, typename VectorCoeffType, typename ParamType>
   void reset_parameter_of_rate(KineticsType<CoeffType,VectorCoeffType> & rate,
                                KineticsModel::Parameters parameter,
-                               const ParamType new_value);
+                               const ParamType new_value, const std::string & unit = "SI");
 
 
 //----------------------------------------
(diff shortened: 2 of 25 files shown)


SylvainPlessis:
Here is some more.  Not yet finished.

This is done by adding a ```set_parameter(KineticsModel::Parameters par, const CoeffType & value)``` to the kinetics models.  One annoying thing is the photochemical model requires a ```CoeffType``` version of this method to be compatible with the rest of the kinetics model, and once I update the photochemical test with it, the other kinetics models will probably require a similar compatibility method.

This raises an interesting question: how do we specify the ```VectorCoeffType``` that is wanted?  Right now this is necessarily a ```std::vector<CoeffType>``` and is set in the ```read_reaction_set_data()``` method.  Solving this is outside the scope of this PR, and I think that compatible methods are enough, but there is some interesting work to be done there.

src/kinetics/include/antioch/reaction.h
```diff
@@ -123,6 +123,12 @@ namespace Antioch
     //! \returns the equation for this reaction.
     std::string equation() const;
 
+    //! \returns the reaction id.
+    std::string id() const;
```
pbauman: Let's have this return a `const std::string&` to avoid the string copy?

src/kinetics/include/antioch/kinetics_parsing.h
```diff
@@ -51,6 +51,15 @@ namespace Antioch
   template<typename CoeffType, typename VectorCoeffType, typename VectorType>
   void reset_rate( KineticsType<CoeffType,VectorCoeffType> & kin, const VectorType & coefs);
 
+  // ParamType is either VectorCoeffType or CoeffType
+  template <typename CoeffType, typename VectorCoeffType, typename ParamType>
+  void reset_parameter_of_rate(KineticsType<CoeffType,VectorCoeffType> & rate,
+                               KineticsModel::Parameters parameter,
+                               const ParamType & new_value);
```
pbauman: Why passing the `ParamType` as a reference? This seems dangerous. Don't we want to pass by value here?

pbauman: Likewise in all the methods where `new_value` is passed by reference. I think we should just be setting the value here.

SylvainPlessis:
Returning a ```std::string``` instead of a const reference was a clear mistake, I changed also the equation getter.  The idea behind the reference for the new value was that in the case of photochemistry, those new values are vectors.  Now this is the setting part, and photochemistry is quite rare, so if you prefer copying, I have no good argument against.

pbauman:
Maybe it makes sense to have a different interface for the photochemistry then? i.e. a `VectorCoeffType` interface where we can pass by reference?

SylvainPlessis:
Woah… This is way too smart for a pre-caffeinated morning...

pbauman:
Maybe not because it's starting to make methods for specific cases which may mean I'm thinking in the wrong direction. I'll be interested in @roystgnr thoughts (when he can come up for air).

SylvainPlessis:
There might be (I want it anyway) in some future a meta-programming stuff to play with this to support vectorization of the ```CoeffType``` type.  The idea is to have a short number of reactions, one per type, and vectorize them to have the full chemical network.  We could potentially gain some good computing time.
So I'm not sure this might go in any wrong/right direction.

Now this vectorization is at the very start of the process, so it still might be worth it to care for it in the meantime.

src/kinetics/include/antioch/kinetics_enum.h
```diff
@@ -68,6 +68,16 @@ namespace Antioch
...
+    enum Parameters{ A = 0,
+                     B,
+                     E,
+                     D,
+                     T_REF,
+                     R_SCALE,
+                     SIGMA,
+                     LAMBDA
+                   };
+
```
roystgnr: Do we have any string_to_enum conversion for these?

SylvainPlessis: I'm not yet at this part, the string will be given at the ```ReactionSet``` level, which will perform both the reaction scanning then the string to enum conversion (using a free method of _string_utils.h_, to be implemented).  I'm leaning towards a switch statement and not building a map, as all the keywords will all be at the same easily `foundable' place.

SylvainPlessis:
This is sorry code, see issue #137 for more details.
I added the compatibility version of the reset methods in all the kinetics models, updated properly the photochemical rate test: it uses both type of resetting the parameter, the rate constant method and the higher level free method which requires all kinetics models objects to be consistent with themselves.

One question for the high level methods: this is resetting one parameter, but it seems that setting it would be better appreciated.

Basically, should the methods at the reaction and reaction set level be called ```reset_parameter_of_reaction``` of ```set_parameter_of_reaction```?

SylvainPlessis:
One important question: what about the unit of the new coefficients?

All informations on the unit is not stored, it is during the reading part, all is converted to SI, then lost.  Should we support custom units in resetting or is it to the user`s responsibility to convert them in SI?

If we support it, how should we declared it and where is it read?  The easiest would be:

```ReactionSet<CoeffType>::set_parameter_of_reaction(const std::string & reaction_id, const std::vector<std::string> & keywords, CoeffType new_value, const std::string & unit = std::string())```

When no unit is provided, SI is assumed.  How do we format the input file?

**Comment by SylvainPlessis** [comment]
Alright first version (unit-wise) of the resetting.  There is no unit testing, SI is therefore assumed.  I'll add the proto unit version by adding the default argument, I'd like to have a little bit of feedback before implementation.

**Comment by SylvainPlessis** [comment]
Alright so I've assumed an input file of the form
```
Antioch/0008/A/0/some unit
Antioch/0008/b/2

Antioch/0014/efficiencies/CH4

Antioch/0022/D/inf/some unit

Antioch/0032/E/0/some unit
Antioch/0032/alpha
```

The unit is optional, is none is provided, SI is assumed.  Antioch takes the unit to be the last (well, actually the following) string given in the ```keywords``` vector of strings.

roystgnr:
So if no unit is specified we assume SI input, if one is we convert?  That's perfect!

SylvainPlessis:
Almost perfect, there is no check whatsoever if there is consistency.

SylvainPlessis:
I think this is the last commits.  Now all the different cases are taken care of, with the subtle Ea case.  The only danger left now is (minus bugs) the no-check.  It's so bad that if you want to reset say the pre-exponential parameter, if you say that the unit is Pascal per litre, it'll just convert it to Pascal per meter cube and be happy with it.

roystgnr:
Looks good, modulo what we just discussed about needing a string-based getter too.  If you want to add get_parameter_of_reaction() to this branch, IMHO we should be ready to merge right afterward.  If you want to merge this right now and put get_parameter_of_reaction() in a separate PR, that's okay too.

SylvainPlessis:
It's better to add it here I think, it should not be overly complicated, and I'll have a nice get/set loop test to add to the _kinetics_regression.C_ test.

I need also to add commentaries, more commentaries, lots of commentaries…

SylvainPlessis:
Last updates:
  - no more troubles between ```CoeffType``` and ```VectorCoeffType```, only ```CoeffType``` parameters are passed and reset.  Thus the fork between photochemistry and the other is done by overloading the methods, see ```Reaction``` and ```KineticsType``` for the details.
  - more comments.
  - the *get* methods

roystgnr:
Regarding the VectorCoeffType case: we decided to handle that the same way we handle the multi-reaction case: with a number as a keyword after the parameter name to index into the conceptually-vector parameter.

I think @pbauman was happy with all the other design decisions after our IM discussion.

The diffs look fantastic to me.  I'll run this through the usual tests now and merge as soon as they pass.

pbauman:
:+1: once @roystgnr is done testing.

PR #138, opened 2015-04-22
Add antioch_do_once for warnings

roystgnr:
Without a do_once test (and some cerr->cout switches), the verbosity in parallel GRINS runs was unbearable.


pbauman:
Thanks for this. Can we set this to 0.3.0-release instead? Then we can merge to master.

roystgnr:
It's just two commits, and when I try to set up a PR it wants to pull in everything else from master too, so let me try just manually cherry-picking.

roystgnr:
Okay, got it into 0.3.0-release.  I'll merge directly from this for the copy in master.

PR #142, opened 2015-04-23
Upgrade/sensitivity

SylvainPlessis:
Same remark for the tag than the get/set PR.  This is a work in progress, but quite big so it comes slowly and progressively.

This is the first implementation of the derivative with respect to parameters for the kinetics.  Except for the photochemistry, all models have the ```sensitivity_*PARAMETER*``` implemented, both with the ```StateType T``` and the ```KineticsConditions cond``` version.  To limit overloading, I've also added a structure, ```return_type<T>```.  This returns ```T``` except if ```T``` is a ```KineticsConditions``` object, in which case it returns a ```StateType```.

So this enables to have two fork methods for every kinetics models:
```
typename return_type<InputType>.::type
  operator(const InputType & T) const {return this->rate(T);}

typename return_type<InputType>::type
  sensitivity(const InputType & T, KineticsModel::Parameters parameter) const;
```

The used methods within theses (```rate``` and ```sensitivity_*PARAMETER*```) are overloaded to enable both versions of the temperature.   A good idea for the future would be to use a similar design for ```derivative```.

### DIFF
--- a/src/core/include/antioch/kinetics_conditions.h
+++ b/src/core/include/antioch/kinetics_conditions.h
@@ -143,6 +143,13 @@ namespace Antioch{
      return _temperature;
   }
 
+ // partial specialization for the return_type
+  template <typename StateType, typename VectorStateType>
+  struct return_type<KineticsConditions<StateType,VectorStateType> >
+  {
+     typedef StateType type;
+  };
+
 } //end namespace Antioch
 
 #endif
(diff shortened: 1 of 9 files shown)


SylvainPlessis:
Alright I'm almost there (tests not considered…).  Everything's in place for kinetics but only with the enum.  I need to know exactly how the strings will be passed to Antioch for this.  This computations will be heavily performed, so passing a string to ```KineticsEvaluator``` to perform string-to-enum operation does not sound a good idea…

PR #144, opened 2015-04-27
Huge bugfix in ASCIIParser (unit change was not updated...)

SylvainPlessis:
This is a very important bug fix: in the ascii parser, I forgot to update the transport mass definition when we switched to SI units (or got mingled in the parsing PR horror).

This corrects this and adds a few optimizations/constant values management upgrade in the viscosities and diffusion kinetics theory models.

This PR should be merged as soon as possible.  It will very probably require to update the values for the new test in the PR#141.

### DIFF
--- a/src/diffusion/include/antioch/molecular_binary_diffusion.h
+++ b/src/diffusion/include/antioch/molecular_binary_diffusion.h
@@ -171,10 +171,10 @@ namespace Antioch{
      _i(si.species()),
      _j(sj.species()),
      _reduced_mass((si.species() == sj.species())?CoeffType(0.5L) * si.M():(si.M() * sj.M()) / (si.M() + sj.M())), // kg/mol
-     _xi((si.polar() == sj.polar())?1L:this->composed_xi(si,sj)),
-     _reduced_LJ_diameter(CoeffType(0.5) * (si.LJ_diameter() + sj.LJ_diameter()) * Units<CoeffType>("ang").get_SI_factor() * ant_pow(_xi,-CoeffType(1)/CoeffType(6))), // 1/2 * (sigma_1 + sigma_2) * xi^(-1/6)
+     _xi((si.polar() == sj.polar())?1:this->composed_xi(si,sj)),
+     _reduced_LJ_diameter(CoeffType(0.5L) * (si.LJ_diameter() + sj.LJ_diameter()) * Units<CoeffType>("ang").get_SI_factor() * ant_pow(_xi,-CoeffType(1)/CoeffType(6))), // 1/2 * (sigma_1 + sigma_2) * xi^(-1/6)
      _reduced_LJ_depth(ant_sqrt(si.LJ_depth() * sj.LJ_depth())  * _xi * _xi), // sqrt(eps_1 * eps_2) * xi^2
-     _reduced_dipole_moment((CoeffType(1e-7) * ant_pow(Constants::light_celerity<CoeffType>(),2)) * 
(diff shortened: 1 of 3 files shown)


src/diffusion/include/antioch/molecular_binary_diffusion.h
```diff
@@ -248,12 +248,11 @@ namespace Antioch{
   {
      _i                     = si.species();
      _j                     = sj.species();
-     _reduced_mass          = (si.species() == sj.species())?si.M():(si.M() * sj.M()) / (si.M() + sj.M());
-     _reduced_dipole_moment = ant_sqrt(si.dipole_moment() * sj.dipole_moment());
+     _reduced_mass          = (si.species() == sj.species())?CoeffType(0.5L) * si.M():(si.M() * sj.M()) / (si.M() + sj.M());
      _xi                    = (si.polar() == sj.polar())?1:this->composed_xi(si,sj);
-     _reduced_LJ_diameter   = CoeffType(0.5) * (si.LJ_diameter() + sj.LJ_diameter()) * Units<CoeffType>("ang").get_SI_factor() * ant_pow(_xi,-CoeffType(1)/CoeffType(6));
+     _reduced_LJ_diameter   = CoeffType(0.5L) * (si.LJ_diameter() + sj.LJ_diameter()) * Units<CoeffType>("ang").get_SI_factor() * ant_pow(_xi,-CoeffType(1)/CoeffType(6));
```
roystgnr: This wasn't necessary; 0.5 is exactly representable as a double

src/diffusion/include/antioch/molecular_binary_diffusion.h
```diff
@@ -184,9 +184,9 @@ namespace Antioch{
         3/16 * sqrt ( 2 * kb / (Navo * pi) )
 */
       _coefficient(CoeffType(0.1875e-25L) * ant_sqrt(2 * Constants::Boltzmann_constant<CoeffType>() * CoeffType(1e25)  /
-                                                            (Constants::Avogadro<CoeffType>() * CoeffType(1e-25) * Constants::pi<CoeffType>())
-                                                          )
-                                         / ( sqrt(_reduced_mass) * (_reduced_LJ_diameter * _reduced_LJ_diameter) )
+                                                      (_reduced_mass * Constants::Avogadro<CoeffType>() * CoeffType(1e-25L) * Constants::pi<CoeffType>())
```
roystgnr: Where did the sqrt() go?

SylvainPlessis: into ant_sqrt()

SylvainPlessis:
Is there a easy way to know which real numbers are exactly representable as double, or do you need to achieve Roy-level understanding of computers?

It would make my life greatly easier if option one is the answer.

Thanks Roy for the remark :smile:

src/diffusion/include/antioch/molecular_binary_diffusion.h
roystgnr: Ah; didn't notice the parenthesis change.

roystgnr:
Rule of thumb for most practical cases: Write it out as a fraction; if the denominator is a power of 2, and neither numerator nor denominator are crazy-huge, then the floating point representation will be exact.  0.5 and 0.875 work, for example; 0.9 doesn't.  The definition of "crazy-huge" depends on 32-bit vs 64-bit etc. but even for float you get to have a numerator in the millions.

roystgnr:
:+1: when @pbauman is happy, by the way

pbauman:
:+1:

PR #148, opened 2015-05-26
Transport Refactor

pbauman:
Apologies in advance for taking a play from @SylvainPlessis's playbook on the length of this PR. I can divide this up into 3 or 4 parts, but it gets worse before it gets better so I at least wanted the finished product to be seen.

This is not yet complete, but preliminary review can begin. The diffs are a little messy, so looking at the final file might help as well.

`PhysicalSet` is now gone. What has replaced it is handling the transport in 3 different container classes: `MixtureViscosity` (this was already there), `MixtureConductivity`, and `MixtureDiffusion`.  (These subclass `MixtureTransport` in order to contain some common code.) These classes contain the species models and provide the API used by `WilkeTransportEvaluator` for evaluating species-level properties. This is effectively what `PhysicalSet` was providing before, but I think this is easier to read and is (at this point) ~1500+ lines shorter. 

`MixtureDiffusion` is still a bit ugly because of the tagging, but some documentation should help. Note that in the history I'd first gone down the path of splitting it into `MixtureSpeciesDiffusion` (for `SpeciesDiffusionBase` models) and MixtureBinaryDiffusion (for `BinaryDiffusionBase` models), but this wasn't going to allow us to reduce the templating to only the species-level model in `WilkeTransportEvaluator`, so I abandoned that approach and decided to have `MixtureDiffusion` handle both types of models. Hence the tagging.

Additionally, for the viscosity and diffusion, I've setup CRTP base classes for the species models to enforce an API on the species evaluations. I can set this up for the conductivity as well, but I'd love some suggestions on the naming (right now there's "with diffusion" and "without diffusion").

At the `WilkeTransportEvaluator` level, we use traits classes to differentiate between calling the correct API depending on the model; currently it's only there for the diffusion, but conductivity is forthcoming. The idea is to hide these details from the user and let the template parameters handle the dispatch. Then, we guard against instantiating with incompatible types as much as possible; if we can't `static_assert`, then we runtime error.

Things still to be done in this PR: documentation pass, restoring compatibility with `WilkeEvaluator`, conductivity traits (for better managing/guarding WilkeTransportEvaluator), and removing `Conditions` template parameters in `WilkeTransportEvaluator`.

[...]

PR #149, opened 2015-06-10
Factor out some implementations

pbauman:
Thus, Antioch will no longer be header only. This is primarily to facilitate a PIMPL idiom for the `GSLSpliner` (forthcoming), but I'm hoping this will have the nice side effect of allowing us to have Travis test with VexCL enabled again since we moved several thousand lines of code into .C files that otherwise won't be compiled in tests (only linked in).

Following this PR, users will need to start linking against `libantioch.so` (or `libantioch.a` or `libantioch.la` as the case may be) thus breaking backward compatibility.

I should note that the last commit (62795c1) is a monster --- it didn't occur to me until after I squashed all the commits together that I could've just left all the `#include`'s alone in the headers and factored out the implementations one at a time and then gone back and fixed the headers in one, smaller commit. If desired, I can go back and do it correctly... Otherwise, history is `make check` clean.

This is built off of #148 so once that's merged I'll rebase here.


pbauman:
I don't know why the commits are in the order they are showing here. I have the correct order in my local `git log`...

pbauman:
@roystgnr I ran this with VexCL and confirmed I didn't hose that. Anything else you think I should run this through before merging? We're effectively breaking backward compatibility here since we'll require linking to `libantioch`.

roystgnr:
Compatibility breakage in the linker arguments doesn't bother me much.  :+1:

ddemidov:
I am trying to look if travis is able to handle vexcl tests now: https://travis-ci.org/ddemidov/antioch/builds/67555027

pbauman:
Thanks @ddemidov Looks like your test never even make it to `make check`... but it also didn't indicate why if failed. Just said job cancelled.

ddemidov:
I amended the commit and cancelled the job. I hope I did everything right this time: https://travis-ci.org/ddemidov/antioch/jobs/67556457

ddemidov:
Looks like there is still not enough memory :(

pbauman:
:( Thanks for taking a look though!

ddemidov:
Is it possible to disable specific tests based on environment variables (There is [TRAVIS=true](http://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables) for example)?

pbauman:
Hmmm, that's a good suggestion. I'm going to open a new ticket to post some ideas for discussion. In the mean time, I'll go ahead and merge this.

PR #150, opened 2015-06-10
Fix for ConstantLewisDiffusivity

pbauman:
Found this during the documentation pass in #148. This is just embarrassing. Fail.

If your Lewis number not equal to 1, then this affects you. We'll have to update all the GRINS/Antioch regression tests.


pbauman:
Forgot to mention, built this off of #149, so will rebase once that one is merged.

pbauman:
Rebased

PR #151, opened 2015-06-10
Use PIMPL idiom to wrap GSLSpliner

pbauman:
Built on #149, so will rebase this if/when that's merged.

The thing that makes this a little gross are the templated member functions in `GSLSpliner`. I had to make a `GSLSplinerShim` class to wrap up the `GSLSplinerImplementation` to actually hide it from the user. Otherwise, GRINS can now compile Antioch (with the 3 lines to link with `libantioch.la`) with GSL enabled in Antioch, without having to check for GSL itself.

Barring bug fixes, release cleanup, etc., I think this finishes off 0.3.0.


pbauman:
Rebased.

@roystgnr would you mind taking a look to make sure this isn't too offensive or there aren't better ways to do this?

roystgnr:
Skimmed the code, still don't understand: why do you need the separate shim?

pbauman:
For things like [this](https://github.com/pbauman/antioch/blob/gsl-pimpl/src/utilities/include/antioch/gsl_spliner.h#L79) function. It's templated on `StateType` which means it has to be inlined (since `StateType` could be an expression). Since it's inlined, then we can't just forward declare the implementation object --- the compiler will want the definition. So then we'd still be including the raw GSL headers. So the shim's definition gets included, but then we can forward declare the actual implementation within the shim and hide the raw GSL.

I tried to explain this in the [documentation](https://github.com/pbauman/antioch/blob/gsl-pimpl/src/utilities/include/antioch/gsl_spliner_shim.h#L43) --- please let me know how I could make it clearer.

pbauman:
Tests with VexCL passing.

roystgnr:
The definition of that function doesn't actually require the definition of the shim, though.  You're passing a reference to GSLShim, which can be done with a forward declaration.

Ah... but you've got other GSLSpliner methods like spline_init which do call GSLSplinerShim methods, which requires a definition of GSLSplinerShim, which means you can't store gsl data structures inside directly, which is why you have three layers and not just two.

On the gripping hand, you *don't* store gsl data structures inside GSLSplinerImplementation, you store pointers to them.  Couldn't you just forward declare gsl_interp_accel and gsl_spline?

pbauman:
> Couldn't you just forward declare gsl_interp_accel and gsl_spline?

Hmm. I thought I'd tried the forward declaration game with GSL before and it wasn't working. I'll try again and if it works, drop the shim.

pbauman:
OK, now I remember. [This](http://stackoverflow.com/questions/7256436/forward-declarations-of-unnamed-struct) has a good explanation.

TL;DR: You can't forward declare an anonymous struct. Here's what's in `gsl_spline.h` for example

```
typedef struct {
  gsl_interp * interp;
  double  * x;
  double  * y;
  size_t  size;
} gsl_spline;
```

Also tried the dirty trick in that linked thread (subclassing the struct), but that's not going to fly because `gsl_spline_alloc` returns a pointer that can't know about our type.

Did I miss something? Any other ideas?

roystgnr:
Sorry about the delay; no other ideas, I'm afraid.  Three layers it is.

roystgnr:
:+1:

PR #158, opened 2015-08-06
Assert that the curve fit requested is not NULL

roystgnr:
This should be slightly more helpful than the segfault that
@beanSnippet found in that case.


pbauman:
Thanks! Feel free to merge when Travis is happy.

Issue #163, opened 2015-08-21
forward and reverse orders


beanSnippet:
I have a chemkin reaction input file with one reaction to test forward and reverse orders. Right now, including the final two lines (or any variation with different orders) makes no difference in the output. Are these currently recognized by Antioch?
```
!
! ONE-RXN TEST MODEL
!
ELEMENTS
H O
END
SPECIES
H2 O2 H2O
END
REACTIONS
2H2 + O2 <=> 2H2O        3.0e13    0.0e-16    0.0e-16
FORD/H2 2.0/O2 1.0/
RORD/H2O 1.0/
END
```

pbauman:
Related to #153? Comments @SylvainPlessis

beanSnippet:
Not sure if this helps, but right now reverse rates look good. It's just
that I try to change the order of forward or backward reactions and nothing
changes.

variscarey:
I don't think the Chemkin parser checks for FORD or RORD in the file
currently(and thus nothing happens).   This didn't come up in testing as
only one of the chemkin inputs we received from the EXACT people had FORD
(and didn't have RORD at all).  There are probably some other gaps in
parsing coverage for some of the more esoteric Chemkin options.


Varis




On Fri, Aug 21, 2015 at 4:15 PM, Rebecca Morrison <notifications@github.com>
wrote:

> Not sure if this helps, but right now reverse rates look good. It's just
> that I try to change the order of forward or backward reactions and nothing
> changes.
>
>
>
> On Friday, August 21, 2015, Paul T. Bauman <notifications@github.com>
> wrote:
>
> > Related to #153 <https://github.com/libantioch/antioch/issues/153>?
> > Comments @SylvainPlessis <https://github.com/SylvainPlessis>
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/libantioch/antioch/issues/163#issuecomment-133572578
> >.
>
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/libantioch/antioch/issues/163#issuecomment-133578016>.
>



-- 
Varis Carey
Assistant Professor
Department of Mathematical and Statistical Sciences
University of Colorado Denver

beanSnippet:
I see. What about in xml?

pbauman:
Thanks @variscarey.

[...]

PR #164, opened 2015-08-24
Add CPPUnit Testing Framework

pbauman:
I've changed my mind about CPPUnit and am very much desiring the ease-of-adding new tests to the testing suite, especially in light of #163 etc. where we'll need to add more coverage.

I'd like to eventually migrate all unit tests (and regression tests?) to this framework. For now, I'm just trying to lay the groundwork to make adding new tests easy.

This is not complete yet for this PR, but the structure is there (using just the Arrhenius rate test) so that we can have a discussion about the setup.

My thinking was as follows:
* There are "standard" tests that don't require any external dependencies (modulo CPPUnit...). There's a program `standard_unit_tests` that will run all these.
* Each of the external dependencies (e.g. Eigen, VexCL, etc.) will have their own program to run. Currently, this PR has the `eigen_unit_tests`.

I've tried to style the tests to be as reusable as I can, but I'm sure there are parts I could do better. E.g. in the Eigen tests, I reuse the `ArrheniusRateVectorTestBase` class. So it's header only and then we just compile the sources needed for each in their own directory, e.g. `standard_unit/arrhenius_rate_valarray_test.C` and `eigen_unit/arrhenius_rate_eigen_test.C`.

The main reason I went this route is that it makes it clear what tests we're skipping and what we're not. This is as opposed to ifdef'ing out the tests and only have one-cppunit-program-to-rule-them-all. Thoughts on this setup?

If folks are OK with this setup, I'll drop the `MetaPhysicL`, `VexCL`, and `GSL` ones in and consider this PR complete for laying the ground work. Then we can migrate the existing tests in our spare time and add new ones to the CPPUnit framework.


pbauman:
Discussion related to separating tests for dependencies happened offline with @roystgnr concluding
> Yeah, that's a fine idea.

So I will proceed with this.

test/standard_unit/arrhenius_rate_vector_test_base.h
```diff
@@ -0,0 +1,123 @@
...
+#include "antioch/physical_constants.h"
+#include "antioch/units.h"
+
+// Base class
+#include "arrhenius_rate_test_helper.h"
+#include "reaction_rate_vector_test_base.h"
+
+template<typename PairScalars>
+class ArrheniusRateVectorTestBase : public ArrheniusRateTestHelper</*Scalar*/typename Antioch::value_type<PairScalars>::type>,
+                                    public ReactionRateVectorBaseTest<Antioch::ArrheniusRate</*Scalar*/typename Antioch::value_type<PairScalars>::type>,PairScalars>
```
dmcdougall: Do you have a class naming inconsistency?  `*TestBase` versus `*BaseTest`.

test/standard_unit/arrhenius_rate_vector_test_base.h
```diff
@@ -0,0 +1,123 @@
...
+// Antioch
+#include "antioch/physical_constants.h"
+#include "antioch/units.h"
+
+// Base class
+#include "arrhenius_rate_test_helper.h"
+#include "reaction_rate_vector_test_base.h"
+
+template<typename PairScalars>
+class ArrheniusRateVectorTestBase : public ArrheniusRateTestHelper</*Scalar*/typename Antioch::value_type<PairScalars>::type>,
```
dmcdougall: It's not clear that `ArreniusRateTestHelper` is actually needed.  Is there a reason against putting all the members of `ArrheniusRateTestHelper` in `ArrheniusRateVectorTestBase`?

dmcdougall:
Can we enable doxygen for the test classes?  It'd be easier to see the class hierarchy that way.

test/standard_unit/arrhenius_rate_vector_test_base.h
pbauman: I reuse `ArreniusRateTestHelper` in defining the non-vector tests [here](https://github.com/pbauman/antioch/blob/cpp-unit/test/standard_unit/arrhenius_rate_test.C#L44). So I think I'll leave it as is.

pbauman:
All the optional library CppUnit structure is there now. In the last two commits, I turned on all the optional libraries when Doxygen gets built and then put all the testing in a separate namespace (`AntiochTesting`). I was hoping the latter might help organize the documentation better, but it didn't help all that much, IMHO. 

Thoughts on that? I was thinking maybe an alternative would be to opt-in at configure time to have the testing documentation built and by default it's not.

Regardless I've addressed the other comments (thanks @dmcdougall) and the testing infrastructure is there so this is ready to go barring comments on the documentation (and anything else that comes up if @roystgnr takes a look).

dmcdougall:
> I turned on all the optional libraries when Doxygen gets built and then put all the testing in a separate namespace (AntiochTesting). I was hoping the latter might help organize the documentation better, but it didn't help all that much, IMHO.

There's a 'namespace' list that Doxygen generates so you can look at just the classes in that namespace.

dmcdougall:
Looks good to me.  Good work @pbauman.

PR #168, opened 2015-09-05
Setup Default Filenames for Install Tree

pbauman:
Before this PR, `DefaultFilename` always pointed toward the source tree of Antioch for the default data files, with nothing able to point to the install tree. This is fine if you always have the source tree lying around, but this is a restrictive assumption. Here, we rename `DefaultFilename` to `DefaultSourceFilename` (with a typedef for `DefaultFilename` for backward compatibility) and add `DefaultInstallFilename`.

I'd talked about this with @roystgnr in IM a week or two ago and we'd settled on something like this PR.


roystgnr:
Looks good.  One of these days we should add hooks for "make install prefix=something_else", though.

pbauman:
Huh. I didn't know that was a thing. Thanks!

Issue #173, opened 2016-01-26
Troe Falloff Reactions and LOW reaction


dsondak:
I would like to provide Antioch with an XML input for a hydrogen-air mechanism.  The Chemkin file contains the following reaction (among many others):

2OH(+M)<=>H2O2(+M)                       7.400E+13    -.370        .00
     LOW  /  2.300E+18    -.900  -1700.00/
     TROE/   .7346   94.00  1756.00  5182.00 /
H2/2.00/ H2O/6.00/ CH4/2.00/ CO/1.50/ CO2/2.00/ C2H6/3.00/ AR/ .70/ 

After discussions with @hovr2pi , I understand the first line to actually be two different reactions:
2OH <=> H2O2
and
2OH + M <=> H2O2 + M.

From my understanding, the first reaction is just an elementary reaction with parameters given by the last three numbers on the first line.

The second reaction is where I am confused.  It does not appear to be the usual ThreeBody reaction (with efficiencies given in the last line).  From discussions with @hovr2pi (and consultations with the Antioch documentation) I understand that Antioch can handle Troe Falloff reactions (with parameters presumably given by the TROE line).  However, I am not sure what to do about the LOW line.  Is it just the parameters for a usual three-body reaction?  Moreover, I simply do not know which one to implement.

Thanks for any help.  Please let me know if you require any clarification.

variscarey:
Technically it is one pressure dependent reaction.  At the low pressure
limit the 3rd body(+M) is required for the reaction.  At high pressure it
is the other (no M) reaction.  The falloff model is the TROE type and as
you state, specifies the type and parameters for the model between the low
and high-pressure limits.

For more details see section 2.4.3.(pressure dependent reactions) in the
Chemkin manual.

Best,
Varis


On Tue, Jan 26, 2016 at 2:14 PM, dsondak <notifications@github.com> wrote:

> I would like to provide Antioch with an XML input for a hydrogen-air
> mechanism The Chemkin file contains the following reaction (among many
> others):
>
> 2OH(+M)<=>H2O2(+M) 7400E+13 -370 00
> LOW / 2300E+18 -900 -170000/
> TROE/ 7346 9400 175600 518200 /
> H2/200/ H2O/600/ CH4/200/ CO/150/ CO2/200/ C2H6/300/ AR/ 70/
>
> After discussions with @hovr2pi <https://github.com/hovr2pi> , I
> understand the first line to actually be two different reactions:
> 2OH <=> H2O2
> and
[...]

PR #174, opened 2016-01-27
Xml parser

SylvainPlessis:
Here is the PR related to issue #173.

It has all the discussed features:

 - if verbose, it says which rate constant is the low pressure and which is the  high pressure limit
 - within a falloff environment, no other value than ``k0`` is allowed for a ``name`` attribute
 - within a falloff environment, no more than one ``name`` attribute is allowed
 - doxygen docs have been updated

Two expected failures tests have been added to test for these two new restrictions, the existing test have been somewhat expanded, meaning some ``name="k0"`` have been put in the first rate constant's block to test for this configuration.

The XFAIL tests are clones of the normal test, with one reaction containing one feature that Antioch now find to be offensive.


SylvainPlessis:
On this, gentlemen, I'm heading for my bed.

See ya!

pbauman:
Closed in favor of #175. Thanks @SylvainPlessis!

PR #189, opened 2016-02-08
NASA Curve Fit Refactor, Part 2

pbauman:
This is the second part of the NASA curve fit refactoring to facilitate updates/fixes to the parsing. This PR updates the internals of `NASA9CurveFit` to only store 9 coefficients. `NASA9CurveFit` now will only accept coefficients in the "nine form".  `CEACurveFit` only accepts coefficients in the "10 form" and will remap the 10 coefficients to the underlying 9. 

Note this preserves backward compatibility with `CEACurveFit` which has been around awhile. `NASA9CurveFit` is new so we don't need to worry about backward compatibility.

With this form for construction, we can easily apply ASCII (`CEACurveFit`) vs. XML parsing (`NASA9CurveFit`).

### DIFF
--- a/src/parsing/include/antioch/ascii_parser.h
+++ b/src/parsing/include/antioch/ascii_parser.h
@@ -122,20 +122,20 @@ namespace Antioch
 ///////////////
 
 //global overload
-        //! reads the thermo, NASA generalist, no templates for virtual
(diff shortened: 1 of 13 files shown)


pbauman:
Threw on a commit that throws an error if thermo parsing is called with the wrong curve fit type.

PR #192, opened 2016-02-23
Have Travis actually use Eigen when testing

pbauman:
It wasn't picking it up before, so let's point Antioch at it during Travis testing. @roystgnr mentioned failures with Eigen, so let's see if this picks them up.


pbauman:
Dammit, looks like Ubuntu Precise uses Eigen 3.0.5-1 and we need >= 3.1.2 (at least, that's what configure says...).

pbauman:
Looks like Trusty has Eigen 3.2.0-8. According to [this page](https://docs.travis-ci.com/user/trusty-ci-environment/), Travis has Trusty in Beta and we can opt-in. I'll push to this branch to at least try it out and then we can discuss whether we want to keep that or not.

pbauman:
OK, I can squash the last two commits into 1. I'd be in favor of merging this, if only because we pick up Eigen testing in Travis. I feel like we're a pretty vanilla build here so that the "beta" part of the Trusty servers won't hit us so much. Thoughts @roystgnr, @SylvainPlessis?

roystgnr:
:+1: after the squash

PR #193, opened 2016-04-14
ArrheniusRate depends on Cf, Ea, not on any "eta"

roystgnr:
This was compiling before, because it was interpreting eta as Ea and
interpreting Ea as R, but it was unnecessarily confusing.


pbauman:
Thanks. Good to merge once Travis is done.

PR #194, opened 2016-04-14
Test AD w.r.t. multiple reaction parameters

roystgnr:
This requires the newest MetaPhysicL (not the newest release, the
stuff I pushed to github 60 seconds ago) to pass.  The test will be
skipped if Antioch is configured without MetaPhysicL.

@hovr2pi, this is the basics of what we want for getting massive parameter sensitivity gradients out of Antioch calls.  The remaining catches are going to be application-side: indexing (we'll need some kind of factory method for associating unique integers to each parameter we try to differentiate against) and setting the independent variable derivatives (this is easy for kinetics parameters, thanks to https://github.com/libantioch/antioch/pull/135, but Varis was going to add the equivalent access for transport and thermo parameters but IIRC never found time).


pbauman:
@roystgnr So is this an AD solution to achieve similar functionality as #142? (We should try and get that merged "soon".)

roystgnr:
It is.  The big hope is that we can thus more easily get gradients w.r.t. lots and lots of reactions without lots and lots of code (in this case, 0 lines in library vs 2000).

pbauman:
I think it will be good to have both as an option, if for no other reason then things like timings etc.

At any rate this is awesome!

PR #197, opened 2016-04-29
0.4.0 release prep

### LINKED ISSUE #110: Fix GSL/build system

pbauman:
Work for 0.4.0 release prep. Cleaned up a couple of warnings, updated, CHANGES, fixed #110 (and a little extra cleanup of gsl_new.m4), and made it so we don't drop the license tool in the distributed tarball (so we get rid of the plethora of output from the license-tool in the builds from tarballs, which would be very confusing to users, probably).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,24 @@
 Version 0.4.0 (currently in development)
+* Antioch no longer header-only
+  - The parsing and a few "core" class implementations
+    have been move to .C files and their symbols are
+    in libantioch. Thus, users must now link against
+    libantioch in their applications.
+* Added temperature exptrapolation capability to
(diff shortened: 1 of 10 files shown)


pbauman:
I should note: make distcheck passes (with and without GSL) and GRINS builds and passes tests with this version of Antioch.

pbauman:
OK, Travis is good. I built/checked this on my laptop and that passed as well. Going ahead and merging and releasing 0.4.0.

PR #198, opened 2016-05-05
Loosen unit test tolerances

roystgnr:
For me this fixes "make check" with a newer gcc and Eigen.

Instead of taking relative tolerances w.r.t. exact reaction rates, we
use tol as a minimum denominator.  This prevents false positives when
a reaction rate of 1e-1600 is off by an order of magnitude.

Slightly bumping up the tolerance in kooij_rate_vec_unit is also
necessary to get float to pass for me.


pbauman:
Thanks!

PR #200, opened 2016-05-06
VexCL/Test Fixes

pbauman:
This gets me passing `make check` with VexCL again.

### DIFF
--- a/test/berthelothercourtessen_rate_vec_unit.C
+++ b/test/berthelothercourtessen_rate_vec_unit.C
@@ -46,12 +46,14 @@
 #include "antioch/eigen_utils_decl.h"
 #include "antioch/metaphysicl_utils_decl.h"
 #include "antioch/valarray_utils_decl.h"
+#include "antioch/vexcl_utils_decl.h"
 
 #include "antioch/berthelothercourtessen_rate.h"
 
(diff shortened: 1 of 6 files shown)


test/berthelothercourtessen_rate_vec_unit.C
```diff
@@ -91,13 +93,13 @@ int check_rate_and_derivative(const PairScalars & rate_exact, const PairScalars
     int return_flag(0);
    for (unsigned int tuple=0; tuple != ANTIOCH_N_TUPLES; ++tuple)
    {
-     if( excessive_difference(T[2*tuple],   rate[2*tuple],     rate_exact[2*tuple],     tol, "rate values") )
+     if( excessive_difference<Scalar>(T[2*tuple],   rate[2*tuple],     rate_exact[2*tuple],     tol, "rate values") )
```
roystgnr: Weird.  Why was this necessary?  Is the vexcl::vec::operator[] return type not the underlying scalar type?

ddemidov: `operator[]` returns a proxy object that takes care of memory transfer.

roystgnr: That makes sense; thanks!

PR #201, opened 2016-05-06
NASAThermo Vectorized Evaluations Fix, More Unit Testing

pbauman:
Builds on #200.

Went down this rabbit hole when removing `CEAThermodynamics` (that's the next PR after this one). The NASAThermoMixture/Evaluators and, therefore, the underlying curve fits, were not exercised in unit tests using vectorized types. This PR makes several fixes to allow that and adds unit testing (in the CppUnit testing framework) for valarray, Eigen, and VexCL types for the NASAThermo. I'll note in the testing, I made sure that the temperatures were such that two different intervals will be need. I'm pretty sure there was a bug in the old `CEAThermodynamics` implementation for vectorized types that we didn't see because we were always in the same temperature interval in the test.

All tests pass for me with Eigen 3.2.8 and VexCL 1.3.3 using GCC 4.9.3. GCC 5.3 has failures in a few of the Eigen tests (nans).

PR #202, opened 2016-05-06
Turn off -Wdeprecated-declarations

pbauman:
This is a Clang option that's churning out so much output that it's
killing our Travis build with Clang, as seen in #201.  I tested with GCC
and no warnings or errors were emitted using this option.


roystgnr:
:+1:

PR #203, opened 2016-05-06
Fix Clang warnings about std::abs

pbauman:
Mentioned in #201. Also, had to adjust tolerances on some of the
tests. For the berthelothercourtessen_rate_unit, I used @roystgnr's
solution in #198 for comparing very small rates.

I just updated the existing tests because I don't have the time at the moment to sink into converting these to the CppUnit framework. But it would be nice to move that direction.


pbauman:
OK, I don't see any of the `std::abs` warning in the Clang build log. Going ahead and merging.

roystgnr:
Thanks for this.

PR #207, opened 2016-06-28
species_viscosities() type should match

roystgnr:
Looks like we broke this in 19553989; I'd just never used that method
before now.


pbauman:
Damn. Thanks for the fix. Shame on me for not having unit tests there for that method (hopefully we can put a dent in increasing coverage when I'm around next week). 

Good to merge after Travis is happy.

PR #208, opened 2016-06-28
Eigen::Array doesn't currently wrap other vectors?

roystgnr:
This looks like a regression in my version of Eigen and not anything
wrong with Antioch, so I'm disabling that test set.

For future reference, the compilation failure was:
```
/usr/include/eigen3/Eigen/src/Core/Array.h: In instantiation of ‘Eigen::Array<_Scalar, _Rows, _Cols, _Options, _MaxRows, _MaxCols>::Array(const T&) [with T = long int; _Scalar = std::valarray<float>; int _Rows = 5; int _Cols = 1; int _Options = 0; int _MaxRows = 5; int _MaxCols = 1]’:
...
/usr/include/eigen3/Eigen/src/Core/Array.h:173:31: error: no matching function for call to ‘Eigen::Array<std::valarray<float>, 5, 1, 0, 5, 1>::_init1(const long int&)’

```

### DIFF
--- a/test/kinetics_regression_vec.C
+++ b/test/kinetics_regression_vec.C
@@ -223,6 +223,12 @@ int vectester(const std::string& input_name,
   int return_flag = 0;
 
 #ifdef ANTIOCH_HAVE_EIGEN
+
+// I can't seem to get fixed Eigen::Array-wrapping-vector-types to
+// compile anymore.  Not even
+// SpeciesVecEigenType v_test(4);
+// We'll disable it for now.
+/*
   {
     typedef Eigen::Array<PairScalars,n_species,1> SpeciesVecEigenType;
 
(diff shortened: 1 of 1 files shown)


pbauman:
What's the version of Eigen for future reference?

roystgnr:
Eigen 3.2.92 according to /usr/include/eigen3/Eigen/src/Core/util/Macros.h; 3.3~beta1-2 according to Ubuntu dpkg.

roystgnr:
Oh, and I presume you're thinking along the same lines as me: if they eventually fix this then we should definitely re-enable these tests but wrap them in EIGEN_VERSION_AT_LEAST-based cpp directives.

pbauman:
> Oh, and I presume you're thinking along the same lines as me: if they eventually fix this then we should definitely re-enable these tests but wrap them in EIGEN_VERSION_AT_LEAST-based cpp directives.

Indeed. :) Thanks!

PR #209, opened 2016-06-28
Better error handling in Blottner parsing

roystgnr:
When we ask for Blottner data for a mixture and we don't have Blottner
data for every species in the mixture, we should get an error message,
not a NULL pointer and impending segfault.


pbauman:
:+1: Thanks!

PR #210, opened 2016-06-28
Add Blottner coefficients interpolating H2O

roystgnr:
I simply solved for three parameters to fit water at 300, 600, and
1200K to NIST viscosity data.

http://webbook.nist.gov/chemistry/fluid/ also has Fluorine, N2O,
Methanol, and some hydrocarbons, if anyone wants to do this again.


pbauman:
Cool. Was the water actually vapor at 300K? Did you use isobaric conditions (e.g. 1 atm pressure)? FWIW, I actually found the original Blottner paper and it looks like they used 1 atm.

roystgnr:
I used 0.1MPa at 600K and 1200K, but viscosity is so insensitive to pressure that I didn't worry about it much.  If someone wants viscosity to more than 4 significant figures then they don't want Blottner in the first place.  At 300K water is liquid anywhere above like 0.003MPa, so I used viscosity from the vapor phase below that.

pbauman:
Cool, thanks. That's all fine, just wanted a log of it and confirm you were doing what I would've done. Also, in [this paper](https://deepblue.lib.umich.edu/bitstream/handle/2027.42/106470/AIAA2013-303.pdf?sequence=1), they used the "pure species" evaluations from ChemKin to fit Blottner profiles (this is our "KineticsTheory"). So, once we pin down what ever bugs we have in our implementation (this is one of the targets I have in mind for next week), we can also use that to generate data for Blottner fits.

Issue #211, opened 2016-06-29
Antioch fails to read gri30.xml


roystgnr:
We get an antioch_not_implemented() error in read_reaction_set_data.C, when the code can't decide what sort of reaction "falloff" is.

In the XML file, there's a `<falloff type="Lindemann"/>` tag, but our code seems to be assuming that we'll have specified that as `type=LindemannFalloff` instead.

SylvainPlessis:
Antioch is much more explicit than the gri30.xml, indeed...

But I wonder why I haven't coded a gri-compatible parser.  The fix is to make ```XMLParser<NumericType>::reaction_chemical_process() ``` aware of this definition, and to make the ```XMLParser<NumericType>::Troe_*_parameter(...)``` compatible with the non-specific block definition.

I'm on it.

pbauman:
We should have a more helpful error message too, e.g. write out what was found (if anything) and then list what is currently understood by Antioch.

roystgnr:
Thanks, @SylvainPlessis !  But don't feel bad if this gets put on the back-burner - this is something I'd like to fix for other users' benefit, not something I need for myself.  Changing the XML file to be Antioch-compatible was pretty trivial, and I'm actually going to be using a different reaction dataset in the end anyway.

pbauman:
I should also note that, if you enjoy pain, I'm pretty sure the ChemKin GRI3 input file is parsed correctly (I'd tested this several months ago).

roystgnr:
@pbauman, we get the "list what is currently understood by Antioch" part right; but the error message omits the "what was found" part, presumably because if you had verbose=true that part was already just printed.

SylvainPlessis:
This is linked to issue #184, du to #172.  This is where I explain that indeed, we were not GRI (Cantera) compatible for falloff.  It was an answer to #163.
Here's the whole history for it.

PR #212 is where I fixed it.

To sum it up, the GRI/Cantera way was not explicit/rigorous enough for me (and I guess I was lazy enough not to look for the order of the Troe parameters), so I used a more explicit description.

roystgnr:
Looks like we're not quite there yet.  #212 doesn't seem to handle three-body reactions?

SylvainPlessis:
Now this one is unexpected...

[...]

PR #212, opened 2016-06-29
xml_parser GRI30 compatible

SylvainPlessis:
not tested yet, it *should* work.

No test yet, but as it's starting to get a little late here, I'm putting it here anyway, so it's at least testable.

So far I've only checked that it doesn't break anything (at least on my computer, let's see what Travis thinks).

Feel free to add the test if you have the time/will.

Edit: And of course, I've forgotten to link it to the issue.

This is the answer to issue #211

### DIFF
--- a/src/parsing/include/antioch/parsing_enum.h
+++ b/src/parsing/include/antioch/parsing_enum.h
@@ -74,6 +74,11 @@ namespace Antioch
                   TROE_F_TSSS
                  };
 
+// GRI30 compatibility
+  enum GRI30Comp{FALLOFF = 0,
+                 TROE
+                };
+
   enum ParsingUnit{
                     MOL_WEIGHT = 0,
                     MASS_ENTHALPY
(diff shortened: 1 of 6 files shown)


SylvainPlessis:
tested and approuved by my computer.  Basically, all the falloff reactions in the file *test/input_files/test_parsing.xml* are doubled to fit the GRI way of falloff, and the test accordingly updated.

Good night!

roystgnr:
Some of these tests are actually failing under the hood, but the failures are being masked because ```if(error > tol) scream_and_die();``` is a no-op for ```error==NaN```.

Reactions 0028bis, 0029bis, 0030bis, 0031bis, 0032bis, 0033bis, and 0034bis are all outputting NaN because alpha, T1, and T3 have been left set to 0, and T2 has been left set to 1e50.

Reaction 0035 is failing, but in that case I'm not sure why.

I can set up a PR that reproduces the failures if it would help.

roystgnr:
Looks like those bis reactions are returning parser->Troe() == false, so they're not querying the Troe parameters in the first place.

roystgnr:
And in turn that's because rate_constant() is looking for a Troe subsection but isn't looking for a falloff subsection of type Troe.

roystgnr:
Merging and rebasing on #228 fixes all the bis reactions for me.  Now to look into 0035.

roystgnr:
0035 is a photochemistry problem, apparently not a parsing problem.

roystgnr:
And 0035 is fixed for me in #229.  I'm *loving* the new regression test; thanks @SylvainPlessis

PR #213, opened 2016-07-01
remove_reference to make C++11 asserts more robust

roystgnr:
This fixes a compile-time issue when I try e.g.
antioch_assert_greater(expression-returning-reference, 0.0) and the
temporary on the right can't be static-cast to the reference type on
the left.


pbauman:
Hey, nice! I bet this is what was going on in libqueso/queso#430 re: compiler errors on assert with C++11. I'll try and port this to QUESO and see if it gets rid of the issues there and then, hopefully, it will be easy to close that out.

roystgnr:
This should fix the "int vs int&" errors; I'm afraid the "string vs char*" errors are a different ballgame though.

pbauman:
Cool, that works because the former were much more numerous than the latter.

Issue #206, opened 2016-06-22
Mandate C++14?


pbauman:
GCC 4.8.2 was C++11 feature complete, so any RHEL 7 based system will natively support C++11. GCC 5 was C++14 complete. My experience has been that the supercomputing centers will deploy a GCC installation if requested, so it shouldn't be a problem.

Intel is more dubious. [C++11 Status](https://software.intel.com/en-us/articles/c0x-features-supported-by-intel-c-compiler) and [C++14 Status](https://software.intel.com/en-us/articles/c14-features-supported-by-intel-c-compiler). Looks like Intel 14+ supports all of the C++11 features we'd want to use and the C++14 auto/decltype feature we'd want.

I think this could be a good move. It would make the code much more readable, I think.

SylvainPlessis:
Because I like things that reference things that reference the this pointer (understand who can), I'll just add this message to say 'hell yeah', and link it to #211.

roystgnr:
Talking to Paul, we're pretty set on requiring C++11, probably after putting out one last C++03 compatible release.  C++14 is still debateable.

roystgnr:
And as of #215 we now require C++11.

roystgnr:
And as of #220 we always enable C++11.

PR #215, opened 2016-07-06
NaN handling, stage 1

roystgnr:
If we have a ridiculously stiff function, we can get overflow of our reaction rates.  This is particularly a concern in reversible reactions where the equilibrium constant underflows at low temperatures - we were actually triggering that in the kinetics_unit tests, but then passing the tests anyway because "if (NaN > tol) error()" doesn't error.

This PR gives us -infinity rather than NaN in overflowing reverse reactions, and adds more testing where appropriate.

I don't think we'll be able to handle infinity smoothly in user code, so stage 2 will be to make that "infinity" user-overrideable.


roystgnr:
I didn't put in any workarounds for missing isnan(), so this PR requires a C++11 feature.  Linking to #206

pbauman:
This looks good. Do we want to individually check for `std::isnan` in the configure checks as part of #206?

roystgnr:
Probably not.  IIRC there are lots of C++03 compilers that include isnan() (whether intentionally or inadvertently via C99 headers) but I've never heard of a C++11 compiler that lacked it.

PR #217, opened 2016-07-06
Reaction::set_maximum_rate()

roystgnr:
Returning infinity was more correct than NaN, but returning huge-but-finite rates might be more practically useful.  *Less* correct, though, so we leave the maximum rate at infinity by default and fob off on the user the responsibility of setting a cap if they need.

Currently I'm only applying this to the reverse reaction rates, which go nuts at sanely low temperatures; if anyone wants to evaluate forward rates at insanely high temperatures then we'll need to add another max  application there.


pbauman:
👍

PR #219, opened 2016-07-06
Gri30

SylvainPlessis:
This should do the trick for the absurdity of ```FalloffThreeBody``` stuff.

I've updated the xml parser, the test is the one described by @roystgnr in #211, namely:

> My personal preference would be to check for \( *\+ *M *\) in the equation, use that to determine the reaction is three-body, but then continue to verify that only three-body reactions have efficiencies.

I don't remember, and  I haven't rechecked yet, but the string ```+ M``` is also used for a type of reaction, we should verify if it could be considered a positive match.

This test **requires** C++11 using ```std::regex```.

The xml parser test has one Troe and one Lindemann falloff three-body more, I don't think it's of any interest to test for all the possible combinations.


SylvainPlessis:
```
terminate called after throwing an instance of 'std::regex_error'
```

I can't reproduce it...  Any idea welcome.  I'll look into it this evening.

Time for :coffee: :grinning:

roystgnr:
I'll give it a try on a couple systems and see if I can reproduce.

roystgnr:
Obsoleted by #228

PR #220, opened 2016-07-07
C++11 fixes

roystgnr:
This also changes the isnan() overloads to has_nan(), something more sensible and less conflicting.

Antioch now once again compiles with the newest gcc and clang.


pbauman:
👍 once Travis is happy.

PR #221, opened 2016-07-07
Distribute antioch.sty

pbauman:
This .sty file encapsulates the extra LaTeX packages we need in
the Doxygen comments. Is now being distributed in the tarball
so we can build the docs from a tarball now.


pbauman:
I did a `make dist`, untarred, and `make doc` and it worked successfully.

PR #218, opened 2016-07-06
Add Examples Directory to Antioch

nicholasmalaya:
This partially hits Numero seis on the todo list, 

<pre>
6. Add examples directory and example programs for usage. Kinetics, thermo, transport.
</pre>

I've added a directory, a makefile.AM, and a single example that builds and runs on my box. I would recommend we merge this guy, and then create a separate ticket that would populate the examples directories with representative cases of the kinetics, thermo and transport API access patterns. Wouldn't mind input from @roystgnr or @pbauman (or David Sondak, is he on github?) about what we might want to add in this case. Or we could look at Cantera and mimic the useful parts of those tests.

PR #227, opened 2016-07-13
Fix make install bug

roystgnr:
This fixes the problem I hit from #218

We'll probably need to rebase #226 on this.

I think "put everything in share/examples" is the most standard way to go, but I could be talked out of that; I'm willing to settle for any install behavior which does not include "error and die".

### DIFF
--- a/examples/Makefile.am
+++ b/examples/Makefile.am
@@ -1,10 +1,5 @@
 EXTRA_DIST = 
 
-examplesdir = $(top_srcdir)/examples
-
-examples_PROGRAMS  =
-examples_PROGRAMS += antioch_init
-
 AM_CPPFLAGS  =
 AM_CPPFLAGS += -I$(top_srcdir)/src/core/include
 AM_CPPFLAGS += -I$(top_srcdir)/src/units/include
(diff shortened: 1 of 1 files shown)


examples/Makefile.am
```diff
@@ -32,7 +27,9 @@ pkginclude_HEADERS =
 #
 # CPP TESTS
 #
+antioch_initdir = $(prefix)/share/examples/antioch_init
 antioch_init_SOURCES = antioch_init.C
+antioch_init_DATA    = ${antioch_init_SOURCES}
 
 #Define examples to actually be run
 TESTS  =
```
dmcdougall: Nuke this and the associated lines below it.  This doesn't belong to the test suite.

examples/Makefile.am
```diff
@@ -32,7 +27,9 @@ pkginclude_HEADERS =
 #
 # CPP TESTS
 #
```
dmcdougall: Can also nuke these three comments.  They're just misleading.

examples/Makefile.am
roystgnr: Leaving them.  More tests is better than fewer, even if we're pretty sure they're redundant.  We probably don't want to establish the *expectation* that examples also be regression tests, but we should leave open the *option*.

examples/Makefile.am
roystgnr: Will do.

examples/Makefile.am
dmcdougall: Makes sense.

dmcdougall:
Thanks @roystgnr, sorry for the hiccup.

PR #229, opened 2016-07-15
Fix NaN errors in kinetics

roystgnr:
This exposes NaNs to more unit tests and assertions, and fixes more NaN errors which cropped up there and in my application code.


pbauman:
Sorry for my sluggishness. This is great. Maybe at some point we should provide a runtime check helper (or something) to check for NANs at the end of the rate calculation so we can detect in opt mode? Just thinking out loud.

Regardless, this looks good to me.

roystgnr:
At the end of the rate calculation it's easy enough for user code to do the checking.  Some kind of "turn on signaling NaN" helper might be a good idea, though.

PR #230, opened 2016-09-29
Bugfix/arrhenius kooij double agent

SylvainPlessis:
Fix to issue #211.

As explained in the issue, the change are to test the rate constant model before checking if a parameter is found, which fixes the double agent Arrhenius falloff problem (first commit), and to keep the Kooij model anyway for the falloff to be sure not to trim the power parameter of the other rate constant (second commit).

Those two situations are added in the _parsing_xml_ test (last commit).

### DIFF
--- a/src/parsing/src/read_reaction_set_data.C
+++ b/src/parsing/src/read_reaction_set_data.C
@@ -359,7 +359,7 @@ namespace Antioch
               (is_k0)?std::cout << "  Low pressure limit rate constant\n":std::cout << "  High pressure limit rate constant\n";
             }
 
-            // pre-exponential
+            // pre-exponential, everyone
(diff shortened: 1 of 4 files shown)


pbauman:
So, this looks OK to me. Was there a reason we didn't merge this right away? I believe @klbudzin hit the issue this fixes recently?

roystgnr:
IIRC it worked for me but I never found time to give it any real review or testing.  I'm okay to merge if you are, though.

pbauman:
It looks like he has some testing in this PR. I'll try and confirm later today and, if satisfied, I'll go ahead and merge.

pbauman:
OK, I pulled this down and ran it. Looks good. I haven't actually tried parsing gri3 kinetics yet, but that can wait for a separate PR.

PR #231, opened 2016-12-12
Derivative of c_p with respect to T

roystgnr:
Only implemented in the NASA curve fits for now.


pbauman:
Sorry for the delay here. Thanks!

roystgnr:
:-P  That was too *little* delay, not too much.  I'd intended to squash those last silly bugfixes before merging.  My fault for not adding a do_not_merge tag.

PR #234, opened 2016-12-19
Add non-const NASAThermoMixture::curve_fit()

roystgnr:
We only had a const accessor, but we need a non-const accessor if we
want to let users change parameters.

Hopefully this will fix the bug @dsondak found in the new
set_curve_fit_coefficient() method.


roystgnr:
Worked for @dsondak so I'm merging right away; it's a few lines which fix an obvious bug.

roystgnr:
On the other hand, the fact that this fix was necessary to begin with does call into question my grasp of the obvious...

PR #236, opened 2016-12-20
Disable Eigen Matrix-of-Array tests

roystgnr:
This fails on multiple gcc versions and -std= options for me, with
what looks like inadequacies internal to Eigen 3.3.1 and 3.2.6, so I
guess it's not supported.


roystgnr:
Merging right away since I'm just turning off a broken test, but if anyone has any idea on how to instead *fix* the broken test I'm all ears.  :-)

Issue #238, opened 2017-03-28
API Consistency Between IdealGasMicroThermo and StatMechThermodynamics


pbauman:
Right now, we don't have complete consistency in the API between these two classes, even though the intent is that they be used interchangeably. For example, cv(), cp(), cv_el(), and probably more, are missing from `IdealGasMicroThermo` but are present in `StatMechThermodynamics`.

I'm thinking we introduce a `MicroThermoBase` CRTP base class and have each of `IdealGasMicroThermo` and `StatMechThermodynamics` appropriately subclass.

roystgnr:
Sounds good to me.

pbauman:
Addressed in #240.

PR #240, opened 2017-04-11
Thermodynamic API Consistency

pbauman:
Addresses #238

This uses a CRTP pattern to define an API for "MacroMicro" thermo objects. That is, objects that provide both "macro" thermo functions (e.g. cp) and "micro" thermo functions (e.g. cv_trans). Doing this also removed some duplicate code from `IdealGasMicroThermo`. It may not be appropriate to have these in a combined object like this, but we can revisit when we want to start thinking about multi-temperature models.

I'd like to merge quickly because we'll consume this in GRINS right away.


roystgnr:
I'm a big fan of the goal of the changes, but I don't have time to go through line-by-line this week.  Could you throw in a few unit tests before we merge?  If not then we can settle for testing via GRINS for the short term.

pbauman:
I've migrated a lot of testing here, but not quite done. Need coverage for `cv_vib` for `StatMechThermodynamics` and `cv_el` for both `IdealGasMicroThermo` (easy) and `StatMechThermodynamics` (more involved). Hopefully will finish this week. The coverage for `cv_vib` in `IdealGasMicroThermo` is good though - I actually test parsing NASA coefficients as the user would do against hand coded functions for the NASA cv curve fits for evaluating `cv_vib`.

And I'm glad I resisted the urge to say "we have already have coverage from old style tests, let's merge" and instead said "let's migrate these tests to CppUnit and make sure we have coverage" since it uncovered #241. Thanks for the prod @roystgnr.

pbauman:
OK, all the methods in `MacroMicroThermoBase` and their implementations are covered by testing. I've also made sure whatever was being covered in the older tests is also covered. As such, I think this is good to merge when Travis is happy.

PR #245, opened 2017-08-10
Adding a test to parse the gri30 xml, this catches the error that was fixed ...

klbudzin:
involving parsing Troe Reactions with species like CH2(s). It reads every reaction in the gri30.xml.


roystgnr:
Great; thanks!  Looks ready to merge unless there are any objections.

pbauman:
Just to clarify, it looks like you're only checking the value of coefficients thermo?

klbudzin:
As checking goes, yes that seems right, I wasn't sure what all I should be doing with it.  But I mainly made it to make sure an error wasn't thrown during the reading of the the reactions of the xml file.

PR #246, opened 2018-04-13
Fix unused variable warning

tradowsk:
This removes an annoying `warning: unused variable ‘n_cols’ [-Wunused-variable]` that I get when building `GRINS` `opt`.


roystgnr:
Thanks!

PR #254, opened 2018-06-26
XML Parser Updates

### LINKED ISSUE #4: Antioch needs to be able to select a reactionData id
Although cantera-style .xml files can have more than one reactionData section, Antioch currently only reads the first section in the file. This doesn't actually affect anything that only has one section, but it would be a good capability to restore for future compatibility.

pbauman:
Closes #4 

XMLParser now has a constructor that takes the `phase` name. Within the `phase` section of the XML file, you can point to a specific `speciesArray` and `reactionArray`, allowing multiple species and reaction sections in the XML file. Previously, we were just grabbing the first one we found. I've preserved that behavior for backward compatibility, but have deprecated it. This will also allow us to parse the active species directly from the XML file (e.g. for use in GRINS).

I also beefed up the testing on the GRI3 file that is shipped with Cantera (and that we ship with Antioch) as part of this effort. So we are testing this new feature and also hitting cross-compatibility with Cantera's format better. (And we do want to support it, their ck2cti utility is really useful as well as ctml_writer for converting their very easy-to-read (and write!) cti format to this XML).

### DIFF
--- a/src/parsing/include/antioch/parser_base.h
+++ b/src/parsing/include/antioch/parser_base.h
@@ -67,7 +67,7 @@ namespace Antioch
   template <typename NumericType>
   class CEAEvaluator;
 
-// micro
+  // micro
   template <typename NumericType>
   class StatMechThermodynamics;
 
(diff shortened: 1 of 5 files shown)


pbauman:
Squashed in missing header for `std::unique_ptr` that local compiler didn't complain about.

PR #255, opened 2018-06-26
Query NASA Curve Fit Type from XML file

### LINKED ISSUE #251: Add method to query thermo type in XML parser
We need a way to determine whether we have NASA7 or NASA9 in the XML file since we need to know that before we parse (because of the templating on the NASACurveFit).

pbauman:
Should be merged after #254.

Closes #251.

Added function that queries the XML file. We just look at the first species and check for the key for NASA7 or NASA9 and error if it's not one of the two. We rely on the subsequent parsing of the thermo during the `read_thermodynamic_data` call to catch inconsistent/mixed use of NASA7 and NASA9. There is unit testing of both NASA7 and NASA9 cases.

### DIFF
--- a/src/parsing/include/antioch/parser_base.h
+++ b/src/parsing/include/antioch/parser_base.h
@@ -67,7 +67,7 @@ namespace Antioch
   template <typename NumericType>
   class CEAEvaluator;
 
-// micro
+  // micro
   template <typename NumericType>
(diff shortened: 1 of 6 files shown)


pbauman:
Will merge this and #254 in the morning if there are no objections.

Issue #260, opened 2018-07-10
Consolidate All Parsing Methods


pbauman:
There's scattered global functions for parsing things, but these should just be base class methods in the ParserBase object and then using virtual functions for implementation specific things. Right now, this is 1. Confusing as hell to figure out what's going on, 2. Makes it harder to update to use new features, e.g. #254 where we can now specify the gas mixture to select what we want if there's multiple sections in the XML file; right now we can't use it for kinetics., 3. Messy.

#250 will be a step in this direction. This will also help with #249.

PR #261, opened 2018-07-10
Update read_reaction_set_data API

pbauman:
This adds a new overload of `read_reaction_set_data` that takes an existing pointer to a `ParserBase` object. Really, we just want to do #260, but I need this right away so that I can leverage the mixture specification in the XML file (which requires the constructor from #254 and is not exposed to the other `read_reaction_set_data` methods). So with this PR, I can construct the XMLParser object myself and just pass to this method. Added unit test of the function that also tests xml file with multiple species and reaction sections in it.

I'd like to merge quickly after a little GRINS testing.


pbauman:
GRINS is good with this PR.

share/antioch_default_files/antioch_default_chemical_mixture.dat
```diff
@@ -80,6 +80,7 @@ O	16.00000	  1.5420000000e7 	1.5 	 0
 O+	15.99945	  9.7560000000e7 	1.5 	 1
 O2	32.00000	  0.000000000000 	2.5 	 0
 O2+	31.99945	  3.6370000000e7 	2.5 	 1
+O3      48.00000          3.0104166666e6        3.0      0
```
roystgnr: Is this also from http://cccbdb.nist.gov/hf0K.asp ?

roystgnr:
:+1: now if you're still consistent with that comment; :+1: after you cite a source otherwise.

pbauman:
lol, just force pushed a small API update to my new API. I'd copy-pasted, but the filename argument in the new function was unnecessary/superfluous. Not sure why compiler didn't throw unused-variable warning.

share/antioch_default_files/antioch_default_chemical_mixture.dat
pbauman: Correct. I should've included in the commit log that the comment exactly as stated is still correct. I'll force push an amendment to the commit log.

pbauman:
OK, force pushed again with amended commit log on the Antioch default data.

PR #267, opened 2018-08-24
Fixing some errors calculating the Diffusion coefficients and thermal conductivity for kinetics theory.

klbudzin:
When comparing Transport values with cantera I noticed a few problems that arose in the calculation of the mixture averaged Diffusion coefficients. First, we are not clipping non-realistic values for mass fractions, so if the user doesn't do this themselves, were throwing out non-physical values. Secondly I converted how we calculate the diffusing mixing rule to use mole fractions instead of mass fractions. For some reason, using mass fractions gives us a different mixture molecular weight than the mole fractions. From my testing, the diffusion coefficients are very sensitive to the mixture molecular weight causing our values to be highly different from the expected values for near pure species mixture concentrations.

I also changed how we calculate the thermal conductivity to match how it is calculated in ("Chemically Reacting Flow" by Kee, Coltrin, and Glarborg, page 519). These values now seem to match to canteras calculated values to at least the third digit. Finally there is also a changed to the unit test that was tripped with the thermal conductivity change.

### DIFF
--- a/src/transport/include/antioch/mixture_averaged_transport_evaluator.h
+++ b/src/transport/include/antioch/mixture_averaged_transport_evaluator.h
@@ -43,6 +43,9 @@
 #include "antioch/diffusion_traits.h"
 #include "antioch/conductivity_traits.h"
 
+// C++
+#include <limits>
+
 namespace Antioch
 {
   //! Compute transport properties using ``mixture averaged" model
@@ -86,6 +89,7 @@ namespace Antioch
     enum DiffusivityType { MOLE_FLUX_MOLE_FRACTION,
                            MASS_FLUX_MASS_FRACTION,
                            MASS_FLUX_MOLE_FRACTION };
+
 
     //! Mixture diffusivities for each species, in [m^2/s]
     /*! Only valid for BinaryDiffusionBase species diffusion models.
@@ -165,6 +169,17 @@ namespace Antioch
                                 const MatrixStateType & D_mat,
                                 DiffusivityType diff_type,
                                 VectorStateType & D_vec ) const;
+
+    //! Compute Perturbe mole fractions
+    /*! To avoid singularities from dividing by very small numbers in the vanishing mass fraction regions, we use the EGLIB approach
+     *  (http://blanche.polytechnique.fr/www.eglib/manual.ps page 5)
+     *  We Use a perturbed mole fraction that ensures that the mole fractions are between 0 and 1.
+     */
+    template <typename VectorStateType>
+    void perturbed_molar_fractions( const ChemicalMixture<CoeffType> & mixture,
+                                    const VectorStateType & mass_fractions,
+                                    VectorStateType & perturbed_mass_fractions) const;
+
 
     const MixtureAveragedTransportMixture<CoeffType>& _mixture;
 
@@ -342,6 +357,8 @@ namespace Antioch
                                              DiffusionTraits<Diff>::is_species_diffusion),
                                             "Incompatible thermal conductivity and diffusion models!" );
 
+
+
     mu_mix = zero_clone(T);
     k_mix  = zero_clone(T);
     StateType k_sum1 = zero_clone(T);
@@ -351,21 +368,25 @@ namespace Antioch
     VectorStateType mu  = zero_clone(mass_fractions);
     VectorStateType k  = zero_clone(mass_fractions);
     VectorStateType chi = zero_clone(mass_fractions);
+    VectorStateType perturbed_mass_fractions = zero_clone(mass_fractions);
+    this->perturbed_molar_fractions( _mixture.chem_mixture(),
+                                     mass_fractions,
+                                     perturbed_mass_fractions);
 
     typedef typename Antioch::rebind<VectorStateType,VectorStateType>::type MatrixStateType;
 
     MatrixStateType mu_mu_sqrt(mu.size());
     Antioch::init_constant(mu_mu_sqrt,mu);
 
-    MatrixStateType D_mat(mass_fractions.size());
+    MatrixStateType D_mat(perturbed_mass_fractions.size());
     init_constant(D_mat,D_vec);
 
 
-    this->compute_mu_chi( T, mass_fractions, mu, chi );
+    this->compute_mu_chi( T, perturbed_mass_fractions, mu, chi );
 
     this->compute_mu_mu_sqrt( mu, mu_mu_sqrt);
 
-    const StateType molar_density = rho / _mixture.chem_mixture().M(mass_fractions); // total molar density
+    const StateType molar_density = rho / _mixture.chem_mixture().M(perturbed_mass_fractions); // total molar density
 
     // If we're using a binary diffusion model, compute D_mat, D_vec now
     if( DiffusionTraits<Diff>::is_binary_diffusion )
@@ -373,7 +394,7 @@ namespace Antioch
         _diffusion.compute_binary_diffusion_matrix(T, molar_density, D_mat);
 
         this->diffusion_mixing_rule<VectorStateType,MatrixStateType>( _mixture.chem_mixture(),
-                                                                      mass_fractions,
+                                                                      perturbed_mass_fractions,
                                                                       D_mat,
                                                                       diff_type,
                                                                       D_vec );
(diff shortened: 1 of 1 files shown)


src/transport/include/antioch/mixture_averaged_transport_evaluator.h
```diff
@@ -579,12 +579,23 @@ namespace Antioch
         }
       case(MASS_FLUX_MASS_FRACTION):
         {
-          VectorStateType molar_fractions = zero_clone(mass_fractions);
-
-          mixture.X(mixture.M(mass_fractions),mass_fractions,molar_fractions);
-
+         
+VectorStateType molar_fractions = zero_clone(mass_fractions);
```
roystgnr: What happened to the indentation here?

src/transport/include/antioch/mixture_averaged_transport_evaluator.h
```diff
@@ -579,12 +579,23 @@ namespace Antioch
...
+	  mixture.X(mixture.M(mass_fractions),mass_fractions,molar_fractions);
+
+          //Clipping unrealistic values, and calculating MW_Mixture through the new mole fractions 
+	  typename value_type<VectorStateType>::type MW_Mixture = zero_clone(mass_fractions[0]);
+	  for(unsigned int s=0; s < D_vec.size(); s++)
+	    {
+	      if(molar_fractions[s] > 1)
+		molar_fractions[s] = 1;
+	      if(molar_fractions[s] < 1e-16)
+		molar_fractions[s] = 1e-16;
```
roystgnr: Should we shout a warning here if we hit these cases?  This seems at first glance like the best thing to do, but imagine e.g. someone trying to take a derivative by central finite differencing around 0; adding clipping under the hood could turn the relative error from O(epsilon^2) to O(.5).

src/transport/include/antioch/mixture_averaged_transport_evaluator.h
```diff
@@ -398,11 +400,12 @@ namespace Antioch
...
           }
 
-        k_mix += k[s]*chi[s]/phi_s;
+        k_sum1 += k[s]*chi[s];
+	k_sum2 += chi[s]/k[s];
         mu_mix += mu[s]*chi[s]/phi_s;
 
       }
-
+    k_mix = .5*(k_sum1+1/k_sum2);
```
roystgnr: We need this better commented.  Of course, we needed the *previous* version much better commented...

Why is this new formula appropriate for k but not mu?

src/transport/include/antioch/mixture_averaged_transport_evaluator.h
pbauman: @klbudzin add the following to your .emacs file in your home directory:

<pre>
(setq-default indent-tabs-mode nil)
(add-hook 'before-save-hook 'delete-trailing-whitespace)
</pre>

or if you're using VIM, ask @roystgnr for the appropriate settings for .vimrc. :)

roystgnr:
Is this going to trip a bunch of GRINS regression tests too?

src/transport/include/antioch/mixture_averaged_transport_evaluator.h
roystgnr: Oh, I love the idea of a ```clip_values = false``` default argument.

src/transport/include/antioch/mixture_averaged_transport_evaluator.h
roystgnr: Ok, works for me then.  Still curious why it would be an improvement for conductivity but not viscosity, though.

pbauman:
> Is this going to trip a bunch of GRINS regression tests too?

At least one. I told him to do this first then we'll update GRINS down stream.

src/transport/include/antioch/mixture_averaged_transport_evaluator.h
klbudzin: I guess I didn't check everything all the way through. My bad, I will ad that to my .emacs file, so hopefully that doesn't happen again.

src/transport/include/antioch/mixture_averaged_transport_evaluator.h
klbudzin: Yeah, most of this was just a quick fix to check that this was what was causing the discrepancies seen between antioch and cantera. The default argument would definitely be better.

src/transport/include/antioch/mixture_averaged_transport_evaluator.h
klbudzin: I'm not sure really, the reacting flow book I obtained it from, basically just stated this formula for the conductivity ontop of the normal viscosity one. I'll look into it more to see if I can obtain a better reasoning.

variscarey: There are a class of empirical averaging formulas(the Kee formula is a version of one of these) that are less accurate(but much cheaper) than the Wilke formula.   There is a long discussion of this in 
A. Ern and V. Giovangigli, Multicomponent Transport Algorithms, Lecture Notes in Physics, New Series Monographs, m 24, (Springer-Verlag, Heidelberg, 1994).

I don't know why Sylvain implemented the Wilke rule for thermal conductivity(I told him to use the Kee formula ala the Chemkin manual for consistency).  Both Chemkin and Cantera use the Kee empirical rule.

src/transport/include/antioch/mixture_averaged_transport_evaluator.h
variscarey: > Secondly I converted how we calculate the diffusing mixing rule to use mole fractions instead of mass fractions. For some reason, using mass fractions gives us a different mixture molecular weight than the mole fractions. From my testing, the diffusion coefficients are very sensitive to the mixture molecular weight causing our values to be highly different from the expected values for near pure species mixture concentrations.

The mole fraction route is the way to go: I asked Sylvain to implement this but he never did(not sure why).  For further stability, Ern & G. recommend the following treatment:

For transport computations, define X_i^tr = X_i + \epsilon(number of species)^{-1} \sum_{species} X_j - X_i), where epsilon is something like machine(double) precision.  This helps with stability in the pure species limit.  It also does not change the sum of the mole fractions(up to rounding, and produces mole fractions between 0 and 1. You then compute an adjusted mass fraction Y_i^tr and transport coefficients using X_i^tr and Y_i^tr.

src/transport/include/antioch/mixture_averaged_transport_evaluator.h
SylvainPlessis: Hello all,

Well I won't be able to remember the whole situation, but it seems to me that there was lots of considerations in this part.  Isn't it also the part where there was an implementation given by a french fortran code?  I remember using their tricks to deal with Titan's high atmosphere (like huge amount of N2, little of CH4 and traces of what we want to follow), and theirs was the only one that would not give the numerical equivalent of kicking the ground in fury, then sit down while breaking in tears...  But I cannot certify it was this exact part of the code.

Might worth a try to look for that code, they might have left a better documentation than me...

Now there is also the argument that I totally prefer rules (like Wilke) than empirical, but that might be because DFT looks sometimes like a bad magic trick to me (Chris can very probably explain what that means).  In that case, my bad, I'm sorry.  There is also the possibility that it was just before I left and I did not do it because I did not find the time before leaving...

src/transport/include/antioch/mixture_averaged_transport_evaluator.h
pbauman: I believe @SylvainPlessis did this in a PR that never got merged and is now stale (#132) :(. @klbudzin can you have a look at what is in #132 and see if it corresponds to what you have here. If not, we should get that integrated.

src/transport/include/antioch/mixture_averaged_transport_evaluator.h
pbauman: It's entirely possible @SylvainPlessis did the Kee formula originally and I changed it because I didn't know any better. (I don't have time to troll through the history). Anyway, I don't want @SylvainPlessis to feel like he's getting dumped on because that's not my intention.

Besides I haven't gotten to speak to either @SylvainPlessis or @variscarey in forever! :(

The French code is EGLib (Ern, Giovangigli) is still available. It's in Fortran. :(

klbudzin:
I've updated a few of the concerns here. I added a separate method to calculate the perturbed mass fractions using the EGlib method, and call it in the mu_k_and_d method before calculating any of the transport properties. I have still yet to add a default argument for the clipping, but I could easily add one to the mu_k_and_d method now. I also added some more documentation in.

PR #269, opened 2019-02-08
Fix redundant warning messages

tradowsk:
Recently did a 520 core `GRINS+Antioch` run and got almost 2000 lines of warnings saying that vibrational levels for O were missing.

This fixes that output mess.


pbauman:
Thanks!

